PR #16, opened 2019-04-24
WIP: Core protocol v3.0 development branch

alimanfoo:
This PR is intended as a focus for work on version 3.0 of the Zarr core protocol. 

To contribute to this PR, please make a PR to the `core-protocol-v3.0-dev` branch. Please **do not** push changes directly to the `core-protocol-v3.0-dev` branch.

Rendering docs from this branch can be [viewed on RTFD](https://zarr-specs.readthedocs.io/en/core-protocol-v3.0-dev/).


alimanfoo:
I added PR #17 with some work towards defining a conceptual model.

jakirkham:
Sorry, I might be missing something here. Does PR ( https://github.com/zarr-developers/zarr-specs/pull/17 ) replace this?

alimanfoo:
> Sorry, I might be missing something here. Does PR ( #17 ) replace this?

No, sorry, #17 was working adding to this. I.e., #17 was a PR on this PR. 

I was thinking that we use this PR and branch (`core-protocol-v3.0-dev`) as the main development branch for work on the v3.0 core protocol, but that we **don't** push directly to this branch, rather we make separate PRs against this branch, e.g., to bring in content for a particular section.

alimanfoo:
I made PR #18 with a straw man for a section on data types, and PR #22 with a straw man for a section on chunk grids.

alimanfoo:
I've added #24 with a straw man for a section on memory layouts.

alimanfoo:
I've added #25 with a straw man for a section on node names.

alimanfoo:
I've added #27 with a suggestion for front matter.

alimanfoo:
I've added #28 with straw man for a section on codecs.

alimanfoo:
I've added #29 with work towards a straw man section on array metadata.

alimanfoo:
I've merged several PRs, latest built document can be viewed here: https://zarr-specs.readthedocs.io/en/core-protocol-v3.0-dev/protocol/core/v3.0.html

alimanfoo:
Here's some open discussion points regarding current state of this branch, feedback on any of these would be welcome:

* Front matter; persistent URLs for the spec; copyright statement; license. 
* Node names, too restrictive? Case insensitive uniqueness of siblings.
* Happy with core data types? Everything else defined via a protocol extension? Extension point mechanism OK?
* Happy with description of chunk grids? OK to only define regular grid here, leave irregular grids and grids which can "grow" in the negative direction to protocol extensions? Extension point mechanism OK?
* Happy with description of memory layouts? Extension point mechanism OK?
* Happy with description of codecs? 
* Happy with group and array metadata? Attributes within the group and array metadata documents, not separate. Zarr format versioning and use of URI.
* Lots to discuss around store interface and storage protocol, see open PR #30.

WardF:
Tagging @DennisheimBigner from Unidata so that he can weigh in if he has any thoughts.  It looks good to me, but Dennis has been giving this a lot more dedicated thought on our end.  @DennisHeimbigner, any thoughts on the changes here, following the discussion we had last week?

docs/protocol/core/v3.0.rst
```diff
@@ -0,0 +1,759 @@
...
+An array is associated with a *data type*. A data type defines the set
+of possible values that the array may contain, and a binary
+representation (i.e., sequence of bytes) for each possible value. For
+example, the little-endian 32-bit signed integer data type defines
+binary representations for all integers in the range −2,147,483,648 to
+2,147,483,647. The core protocol only considers a limited set of data
+types, but protocol extensions may define other data types.
+
+An array is divided into a set of *chunks*, where each chunk is a
+hyperrectangle defined by a tuple of intervals, one for each dimension
```
ttung: are there restrictions on the chunk sizes?

docs/protocol/core/v3.0.rst
```diff
@@ -0,0 +1,759 @@
...
+held in a *store*. To enable a variety of different store types to be
+used, the core protocol defines a simple *store interface* which is a
+common set of operations that a store must provide.
+
+
+Node names
+----------
+
+Except for the root node, each node in a hierarchy must have a name,
+which is a string of ASCII characters. To ensure consistent behaviour
```
ttung: 7-bit ASCII?

docs/protocol/core/v3.0.rst
```diff
@@ -0,0 +1,759 @@
...
+* must consist only of characters in the sets `a-z`, `A-Z`, `0-9`,
+  `-_.`
+
+* must not be a string composed only of period characters, e.g. "." or
+  ".."
+
+* must be at most 255 characters long
+
+Note that node names are used to form storage keys, and that some
+storage systems will perform a case-insensitive comparison of storage
```
ttung: doesn't that mean the storage system is required to perform a case-insensitive comparison of storage keys, in order to have consistent behavior between systems?

docs/protocol/core/v3.0.rst
alimanfoo: I don't think so. Some compression codecs may have a limit on chunk size, but that's a restriction of a specific codec.

docs/protocol/core/v3.0.rst
alimanfoo: Maybe we can avoid mentioning encoding here at all? I.e., just express what characters are allowed, but leave encoding as an implementation detail that's behind the store interface?

docs/protocol/core/v3.0.rst
alimanfoo: This is awkward, it's in there because of mac file systems (HFS+, APFS) being case insensitive by default. With the current zarr python implementation we've seen specific issues of bugs being due to two sibling arrays getting created with names like "foo" and "FOO", with data from one ending up in the other. I was trying to come up with something to address this, but it's possibly not the right approach currently.

Note that IIUC mac file systems do case insensitive matching, but do preserve the original case of the file as created. So, e.g., opening "foo.txt" or "FOO.TXT" will open the same file, but if you created it originally as "FOO.TXT" then that's what you'll see when listing a directory.

alimanfoo:
I've made a PR (https://github.com/zarr-developers/zarr-specs/pull/35) to illustrate how extension data types could be defined and used together with the 3.0 core protocol. This is just meant as an illustration of the data type extension mechanism.

alimanfoo:
OK, here is an update on the current status of this development branch. All of the below is open for discussion and revision, I'm still considering this as straw man work to provide food for thought.

The current state of this branch can be viewed as built documentation [here](https://zarr-specs.readthedocs.io/en/core-protocol-v3.0-dev/).

* I've done some editing to the [concepts and terminology section](https://zarr-specs.readthedocs.io/en/core-protocol-v3.0-dev/protocol/core/v3.0.html#concepts-and-terminology) at the beginning to give it more structure and enable it to be used as a definition list that other sections can link to whenever words with special meanings like "array" or "codec" are used.

* In the interests of being totally minimal, I have pulled the complex number data types out of the core protocol and put them into a [protocol extension](https://zarr-specs.readthedocs.io/en/core-protocol-v3.0-dev/protocol/extensions/complex-dtypes/v1.0.html) (currently just a placeholder, content to be written). Happy to reverse this if objections.

* I have removed support for multiple codecs from the core protocol as suggested in #38. Now the core protocol only defines a single "compressor" codec. I've added a placeholder for a [filters protocol extension](https://zarr-specs.readthedocs.io/en/core-protocol-v3.0-dev/protocol/extensions/filters/v1.0.html) to define support for one or more codecs to be applied during chunk encoding prior to the compressor codec.

* I have added a mechanism that opens the door for the use of different encoding mechanisms for array and group metadata, i.e., other than JSON, as discussed in #37. JSON is still the only and canonical format defined in the core protocol, but other formats could be defined and implemented as extensions. Allowing for different metadata encodings means also providing a mechanism for an implementation to discover what encoding has been used, and so to solve this I've added the concept of a [bootstrap metadata document](https://zarr-specs.readthedocs.io/en/core-protocol-v3.0-dev/protocol/core/v3.0.html#bootstrap-metadata) where the metadata encoding can be specified. The idea is that an implementation would first read the bootstrap metadata, then figure out how to read other metadata. "Bootstrap metadata" doesn't feel like a very good name for it, so if anyone has a better name then please suggest. Again if objections or feelings that this is too complicated, feel free to push back on this.

* User-defined metadata (a.k.a. attributes) are now included within the array and group metadata documents, i.e., not as separate documents (although extensions could still use separate documents for specific purposes should they want to).

* The sections on [stores](https://zarr-specs.readthedocs.io/en/core-protocol-v3.0-dev/protocol/core/v3.0.html#id11) and [storage protocol](https://zarr-specs.readthedocs.io/en/core-protocol-v3.0-dev/protocol/core/v3.0.html#storage-protocol) are really critical and need some careful review and discussion. What's currently there is one possible solution, but there could be others.

Any/all comments welcome as always.

alimanfoo:
Apologies for cross-posting this in a few different places, but I finished up a [slide deck with an overview of current status of the v3 design work](https://zarr-developers.github.io/slides/v3-update-20190619.html). 

I think I've gone about as far as I can with v3 work without some broader discussion and reflection, so if you have a chance please do take a look at the slides and share any thoughts.

docs/protocol/core/v3.0.rst
```diff
@@ -0,0 +1,1164 @@
...
+    name_ of each ancestor node separated by the "/" character,
+    followed by the name_ of the node itself. For example, the path
+    "/foo/bar" identifies a node named "bar", whose parent is named
+    "foo", whose parent is the root of the hierarchy. The path "/"
+    identifies the root node.
+
+.. _dimension:
+.. _dimensions:
+
+*Dimension*
```
shoyer: I am slightly concerned that this may cause confusion with extensions of the zarr spec for dealing with named dimensions (I.e., netcdf). The term "dimension" in netcdf refers specifically to a named dimension (e.g., "time"), which may be shared between multiple arrays.

alimanfoo: Ah that's good to know. Not sure how best to resolve that, this use of "dimension" as in "multi-dimensional array" is also very common (e.g., [numpy](https://docs.scipy.org/doc/numpy-1.16.1/reference/arrays.ndarray.html)). Maybe a clarifying note here to say this is not a named dimension a la netcdf?

docs/protocol/extensions/datetime-dtypes/v1.0.rst
```diff
@@ -0,0 +1,156 @@
...
+because different months have different numbers of days.
+
+
+Units
+=====
+
+====  ============  ====================  ======================
+Code  Meaning       Time span (relative)  Time span (absolute)
+====  ============  ====================  ======================
+Y     year          +/- 9.2e18 years      [9.2e18 BC, 9.2e18 AD]
```
clbarnes: Very minor point: BCE/CE may be preferable over BC/AD.

jeffdlb:
I read with interest the Zarr 3.0 slide deck at https://zarr-developers.github.io/slides/v3-update-20190619.html. Here at NCAR we are currently using Zarr 2.0 for some climate model outputs on AWS (https://doi.org/10.26024/wt24-5j82). I have a couple of cosmetic suggestions you might wish to consider for v.3:

(1) Including a traditional filename extension on the metadata files -- e.g., zarray.json instead of .zarray, zattrs.json instead of .zattrs. It wouldn't make a functional difference, but it's helpful when you're looking at the contents of your S3 bucket and just want to download and inspect one of those files. Could also be useful in future if non-JSON encodings are offered.

(2) Padding the integers used in the chunk object IDs. When there are >10 chunks in any given dimension, the usual alphanumeric sort order gives 0.0, 0.1, 0.10, ..., 0.19, 0.2, ..., 10.0, ..., 2.0, etc. It would be nice to instead have 00.00, 00.01, 00.02, etc. Again, not really a functional impact but just aesthetically more appealing to have the objects listed by (s3)ls in the same order as the actual data.

I realize the idea is that you would normally only interact with the data via the API, but when you look under the hood it's good for things to be nicely arranged.

Cheers,
Jeff de La Beaujardière, NCAR

alimanfoo:
Hi @jeffdlb, I'm realising that I never thanked you for your comments, apologies. Some discussion...

> (1) Including a traditional filename extension on the metadata files -- e.g., zarray.json instead of .zarray, zattrs.json instead of .zattrs. It wouldn't make a functional difference, but it's helpful when you're looking at the contents of your S3 bucket and just want to download and inspect one of those files. Could also be useful in future if non-JSON encodings are offered.

In the [current v3 protocol draft](https://zarr-specs.readthedocs.io/en/core-protocol-v3.0-dev/protocol/core/v3.0.html#storage-keys) I used "zarr.json" as the key for the [bootstrap metadata document](https://zarr-specs.readthedocs.io/en/core-protocol-v3.0-dev/protocol/core/v3.0.html#bootstrap-metadata), which will always be in JSON format. 

I left off any format-specific extension for the array and group keys. E.g., group keys are like "meta/root/foo/bar.group" and array keys are like "meta/root/foo/baz.array". The reasoning there was that extensions might be defined which use something other than JSON encoding for metadata (e.g., flatbuffers). I thought it would be simpler and more flexible to keep the array and group keys independent of the encoding format. Which encoding format is used for array and group metadata would be given in the bootstrap metadata document.

But I see your point that it might be useful to have an extension for the array and group metadata keys so if you're downloading a file from S3 for manual inspection it might be useful to have an extension, so you know what you're dealing with without having to look up the format via the bootstrap metadata document. 

> (2) Padding the integers used in the chunk object IDs. When there are >10 chunks in any given dimension, the usual alphanumeric sort order gives 0.0, 0.1, 0.10, ..., 0.19, 0.2, ..., 10.0, ..., 2.0, etc. It would be nice to instead have 00.00, 00.01, 00.02, etc. Again, not really a functional impact but just aesthetically more appealing to have the objects listed by (s3)ls in the same order as the actual data.

I agree this would be very convenient when viewing in lexical order. The difficulty here is that you can't know up front how many zeros to pad, because an array can be extended along any dimension.
 
> I realize the idea is that you would normally only interact with the data via the API, but when you look under the hood it's good for things to be nicely arranged.

This is important I think, making it easy to work with zarr data and metadata via generic file system and cloud tools is really useful, makes the format very hackable. Hard to strike a balance, discussion very welcome.

jeffdlb:
@alimanfoo Thanks for your Mar 27 reply. 

Regarding filename extensions such as .json on .zarray: I continue to think that they would be useful. Also, a HEAD request on one of our .zarray objects on AWS S3 reveals that ContentType=binary/octet-stream instead of application/json.
I don't know if this is something we should have controlled during upload, but I do know it is not optimal or correct (indeed, binary/octet-stream is not even an IANA-approved MIME type; should be application/octet-stream). By the way, it might be nice if the actual data chunks had a Zarr-specific MIME type such as application/vnd.zarr.chunk.

Regarding zero-padding, the number of digits could be configurable with some appropriate default like three. We have been aiming for ~100 MB chunks, so with 000-999 on each axis we could handle a 100 TB dataset.

alimanfoo:
Hi @jeffdlb,

> Regarding filename extensions such as .json on .zarray: I continue to think that they would be useful. Also, a HEAD request on one of our .zarray objects on AWS S3 reveals that ContentType=binary/octet-stream instead of application/json.
> I don't know if this is something we should have controlled during upload, but I do know it is not optimal or correct (indeed, binary/octet-stream is not even an IANA-approved MIME type; should be application/octet-stream). By the way, it might be nice if the actual data chunks had a Zarr-specific MIME type such as application/vnd.zarr.chunk.

I definitely see the point. The slight tension is that I'd like to generalise in the v3 spec to allow for other metadata encodings to be used, which would mean the suffix (".json") would have to be specified somehow rather than hard-coded. All doable, happy to consider, raised https://github.com/zarr-developers/zarr-specs/issues/63 to track.

For chunks it is easier, these could all have a fixed suffix. I don't know how useful it would be to have a mime-type, given that the arrays cannot be interpreted without also reading the array metadata.

> Regarding zero-padding, the number of digits could be configurable with some appropriate default like two or three. We have been aiming for ~100 MB chunks, so with 00-99 on each axis we could handle a 100 TB dataset; with 000-999 a 100 PB dataset.

I can see this works for specific cases, but it is hard to generalise. Some use cases might use large numbers of very small chunks. Perhaps it could be a configuration option, but I'm hesitant to add complexity to spec and implementations unless it's really worthwhile. Raised https://github.com/zarr-developers/zarr-specs/issues/64 to track.

rabernat:
> Also, a HEAD request on one of our .zarray objects on AWS S3 reveals that ContentType=binary/octet-stream instead of application/json.

Content type is determined by the uploader: https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObject.html

When writing or copying zarr to s3, you could specify a `Content-Type: application/json` header for those files.

alimanfoo:
> > Also, a HEAD request on one of our .zarray objects on AWS S3 reveals that ContentType=binary/octet-stream instead of application/json.
> 
> Content type is determined by the uploader: https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObject.html
> 
> When writing or copying zarr to s3, you could specify a `Content-Type: application/json` header for those files.

Thanks @rabernat, that's useful to know.

joshmoore:
> https://github.com/zarr-developers/zarr-specs/pull/16#issuecomment-622440380 I definitely see the point. The slight tension is that I'd like to generalise in the v3 spec to allow for other metadata formats to be used, which would mean the suffix (".json") would have to be specified somehow rather than hard-coded.

A somewhat related report from the field: in trying to make images in Zarr compatible with existing tools in our domain (e.g. [[ImageJ](http://imagej.net)) we realized that most of the current file choosers disallow choosing directories and hidden files, effectively leaving no way open a Zarr fileset. The prospect of having changing array/group/attribute file names or even file _endings_ could make this even more difficult.

jeffdlb:
> A somewhat related report from the field: in trying to make images in Zarr compatible with existing tools in our domain (e.g. [ImageJ](http://imagej.net)) we realized that most of the current file choosers disallow choosing directories and hidden files, effectively leaving no way open a Zarr fileset. The prospect of having changing array/group/attribute file names or even file _endings_ could make this even more difficult.

@joshmoore Thanks for your comment and for noting this interesting issue. The problem of hidden files suggests to me that the solution is renaming .zattrs, .zgroup, .zmetadata to zattrs.json, zgroup.json, zmetadata.json. 

> > [#16 (comment)](https://github.com/zarr-developers/zarr-specs/pull/16#issuecomment-622440380) I definitely see the point. The slight tension is that I'd like to generalise in the v3 spec to allow for other metadata formats to be used, which would mean the suffix (".json") would have to be specified somehow rather than hard-coded.
> 

@alimanfoo Sorry if I'm being thick, but I don't understand why adding a format-specific suffix would make it harder to enable different metadata formats. Indeed, it would seem to be an essential element of allowing metadata in multiple formats -- e.g., offering both zmetadata.json and zmetadata.xml for the same dataset.

alimanfoo:
> @alimanfoo Sorry if I'm being thick, but I don't understand why adding a format-specific suffix would make it harder to enable different metadata formats. Indeed, it would seem to be an essential element of allowing metadata in multiple formats -- e.g., offering both zmetadata.json and zmetadata.xml for the same dataset.

No apologies necessary. To unpack my assumptions, in the design 

> > A somewhat related report from the field: in trying to make images in Zarr compatible with existing tools in our domain (e.g. [ImageJ](http://imagej.net)) we realized that most of the current file choosers disallow choosing directories and hidden files, effectively leaving no way open a Zarr fileset. The prospect of having changing array/group/attribute file names or even file _endings_ could make this even more difficult.
> 
> @joshmoore Thanks for your comment and for noting this interesting issue. The problem of hidden files suggests to me that the solution is renaming .zattrs, .zgroup, .zmetadata to zattrs.json, zgroup.json, zmetadata.json.
> 
> > >

FWIW the [v3 core protocol spec](https://zarr-specs.readthedocs.io/en/core-protocol-v3.0-dev/protocol/core/v3.0.html#storage-protocol) currently proposes {name}.array and {name}.group to form metadata keys, where {name} is the array or group node name.

alimanfoo:
> @alimanfoo Sorry if I'm being thick, but I don't understand why adding a format-specific suffix would make it harder to enable different metadata formats. Indeed, it would seem to be an essential element of allowing metadata in multiple formats -- e.g., offering both zmetadata.json and zmetadata.xml for the same dataset.

No apologies necessary, there is a lot to unpack here.

FWIW in my work on designing the v3 protocol I have assumed that JSON is the default metadata format, but that a protocol extension might want to override this and use a different format (e.g., flatbuffers). 

In order for a zarr implementation to be able to read a zarr hierarchy, it needs some way to discover what metadata format is in use. Currently in the v3 core protocol draft this is achieved by stating within the [entry point metadata document](https://zarr-specs.readthedocs.io/en/core-protocol-v3.0-dev/protocol/core/v3.0.html#entry-point-metadata) which metadata format is in use, via the `metadata_encoding` property, whose value is a mime-type (e.g., "application/json").

In other words, a zarr implementation starts by opening the entry point metadata document, looks up the `metadata_encoding` property, and checks if it supports the given format. Presumably most implementations will just support "application/json", in which case if another format is given, the implementation will raise a suitable error and stop processing.

If users would prefer a suffix like ".json" to be added to the array and group metadata keys, then a zarr implementation will need some way of discovering this too. 

Perhaps it would be OK to say in the core protocol spec that if the metadata encoding is "application/json" then an implementation must add ".json" as a suffix onto all metadata keys. Any protocol extension that defines a different metadata format would then need to also define a fixed metadata key suffix, and would need to communicate that somehow to implementers.

Alternatively, the metadata key suffix could be given in the entry point metadata document, e.g., via a new property called something like `metadata_key_suffix`. Then nothing is hard-coded, and a zarr implementation would have a general way to discover what suffix is in use.

So it's all doable, it just boils down to some fine-grained decisions about who has responsibility for communicating what suffix to use. I.e., is the suffix something that gets hard-coded into specs and implementations, or is there some more flexible mechanism for discovering what suffix is in use? 

(Btw I have never heard of any requirements to provide metadata for the same hierarchy in multiple formats, so I am assuming that we only need to support one format at a time.)

DennisHeimbigner:
WRT to the metadata object names, It is important
that metadata files can be distinguished, by name., from
user-define names (of e.g. groups and arrays). I might suggest
that metadata object names have the prefix '_.zXXX'

docs/protocol/core/v3.0.rst
```diff
@@ -64,29 +64,104 @@ stores, in particular we want to achieve the following:
...
+This section is informative and is present to help the reader familiar  with
+previous version of zarr to find and understand the differences and the reasons
+behind them as well as guide the contributor during the draft and review
+period.
+
+Better suitability for HPC file systems and network stores
+----------------------------------------------------------
+
+One goal of the spec v3 is to have a design that minimized the number of
+round-trip operations that must done in order to understand the structure of a
```
davidbrochart, line 91:
```suggestion
round-trip operations that must be done in order to understand the structure of a
```

docs/protocol/core/v3.0.rst
```diff
@@ -64,29 +64,104 @@ stores, in particular we want to achieve the following:
...
 
 This means that a store cannot be opened at an arbitrary point, but needs to be
 opened at the root. User facing convenience functions could walk a given
 hierarchy and return a sub-group, but this is not part of the API.
 
+Goal and Non-Goal of v3 spec with respect to v2 spec
+====================================================
+
+This section is informative and is present to help the reader familiar  with
+previous version of zarr to find and understand the differences and the reasons
```
davidbrochart, line 83:
```suggestion
previous versions of zarr to find and understand the differences and the reasons
```

joshmoore:
See open comments on #135

jakirkham:
I wonder if it would make more sense to merge this at this point as opposed to merging more PRs into this one. Understand there may still be rough edges and things we need to sort out still, but am just thinking about this practically. Please feel free to push back on this suggestion though

@joshmoore @rabernat @alimanfoo, thoughts? 🙂

jbms:
This is just about renaming the `core-protocol-v3.0-dev` branch to `master`, right?

That seems fine since the master branch is currently unused.

However, to avoid confusion it might be helpful to clarify in the "About" that this is about zarr v3 only, not zarr v2.

joshmoore:
> However, to avoid confusion it might be helpful to clarify in the "About" that this is about zarr v3 only, not zarr v2.

:+1: (Could even add a section on v2 with links, etc. to help people along)

Waiting a day or two more for final thoughts then this can go in. readthedocs config will need updating at that same time.

joshmoore:
As mentioned. Merging and I'll tidy up the RTD configuration. All future branches should be against the mainline.

PR #17, opened 2019-04-24
Core protocol v3.0 - conceptual model

alimanfoo:
This PR has some initial drafting of text describing a conceptual model for Zarr, which hopefully helps to provide a foundation for everything else to go into the spec. Work towards #16.

### DIFF
--- a/docs/protocol/core/v3.0.rst
+++ b/docs/protocol/core/v3.0.rst
@@ -1,3 +1,45 @@
 Zarr core protocol version 3.0
 ==============================
 
+Conceptual model
+================
+
+A Zarr *hierarchy* is a tree structure, where each node in the tree is
+either a *group* or an *array*. Group nodes may have children
+but array nodes may not.
+
+Each node in a hierarchy has a *name* which is a string of ASCII
+characters with some additional constraints (see below). Two sibling
+nodes cannot have the same name. The root node does not have a
+name.
+
+Each node in a hierarchy has a *path* which uniquely identifies all
+nodes in the hierarchy. The path is formed by joining together the "/"
+character, followed by the names of all ancestor nodes separated by
+the "/" character, followed by the name of the node itself. For
+example, the path "/foo/bar" identifies a node named "bar", whose
+parent is named "foo", whose parent is the root of the hierarchy. The
+string "/" identifies the root node.
+
+An array has a fixed number of *dimensions*. Each dimension has an
+integer length. In the simplest case, the lengths of all dimensions
+are finite. However, a dimension can also be of infinite length, or
+variable length.
+
+The *shape* of an array is the tuple of dimension lengths. For
+example, if an array has 2 dimensions, where the length of the first
+dimension is 100 and the length of the second dimension is 20, then
+the shape of the array is (100, 20).
+
+An array contains zero or more *elements*. Each element can be
+identified by a tuple of coordinates, one for each dimension of the
+array. If all dimensions of an array have finite length, then the
+number of elements in the array is given by the product of the
+dimension lengths.
+
+TODO data type
+
+TODO chunks, chunk shape, chunk grid
+
+TODO additional constraint on names
+
--- a/docs/protocol/core/v3.0.rst
+++ b/docs/protocol/core/v3.0.rst
@@ -13,13 +13,13 @@ characters with some additional constraints (see below). Two sibling
 nodes cannot have the same name. The root node does not have a
 name.
 
-Each node in a hierarchy has a *path* which uniquely identifies all
-nodes in the hierarchy. The path is formed by joining together the "/"
-character, followed by the names of all ancestor nodes separated by
-the "/" character, followed by the name of the node itself. For
-example, the path "/foo/bar" identifies a node named "bar", whose
-parent is named "foo", whose parent is the root of the hierarchy. The
-string "/" identifies the root node.
+Each node in a hierarchy has a *path* which uniquely identifies that
+node and defines its location within the hierarchy. The path is formed 
+by joining together the "/" character, followed by the names of all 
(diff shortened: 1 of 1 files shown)


alimanfoo:
Oops, I meant to create this PR in my own fork, but got a bit confused while using the github online editor and created it within zarr-developers instead. I'll leave this PR as-is for now to avoid confusion but will try to use my own fork for other PRs.

docs/protocol/core/v3.0.rst
```diff
@@ -1,3 +1,152 @@
...
+hierarchy are held in a *store*. To enable a variety of different
+store types to be used, the core protocol defines a simple *store
+interface* which is a common set of operations that a store must
+provide.
+
+
+Node names
+----------
+
+TODO define constraints on node names
```
alimanfoo: N.B., I don't intend to address this TODO or any below in this PR, just adding them as placeholders for a possible structure for other sections.

alimanfoo:
OK, this is now a reasonably complete first pass. Comments welcome.

docs/protocol/core/v3.0.rst
```diff
@@ -1,3 +1,152 @@
...
+
+Each node in a hierarchy has a *path* which uniquely identifies that
+node and defines its location within the hierarchy. The path is formed 
+by joining together the "/" character, followed by the names of all 
+ancestor nodes separated by the "/" character, followed by the name of 
+the node itself. For example, the path "/foo/bar" identifies a node 
+named "bar", whose parent is named "foo", whose parent is the root of 
+the hierarchy. The string "/" identifies the root node.
+
+An array has a fixed number of *dimensions*. Each dimension has an
```
meggart: Is this fixed number >0, i.e. are zero-rank arrays allowed?

alimanfoo: I would say zero-rank arrays should be allowed in the core protocol. I.e., change this sentence to:

```suggestion
An array has a fixed number of zero or more *dimensions*. Each dimension has an
```

However, happy to discuss either way.

constantinpape: Is there a good reason to support rank zero?
I can't think of a use-case right now.

alimanfoo: I believe it's used for scalar values. We added support for this in the python implementation - https://github.com/zarr-developers/zarr/pull/161. 

(Note there was some mixing of discussion there as we also added support for arrays with one or more zero-length dimensions at around the same time and the two requirements got a bit confused, see also - https://github.com/zarr-developers/zarr/pull/160.)

constantinpape: That makes sense. Thanks for the clarification.

alimanfoo:
In the interests of having a straw man and a rough framework for further work, if no objections I'd like to merge this PR which is going into the core-protocol-v3.0-dev branch. I would certainly expect us to revisit some or all of this and so this is not putting anything in stone.

jakirkham:
Not to slow things down here, but do we want to consider soft/hard links? Mentioning this as that might move us away from trees and towards graphs.

alimanfoo:
> Not to slow things down here, but do we want to consider soft/hard links? Mentioning this as that might move us away from trees and towards graphs.

Good question. I was actually thinking that soft/hard links would be something that could be tackled in a protocol extension, but very happy to discuss and revisit.

In general I'm thinking there are a number of features that could be handled as protocol extensions, which would give us a mechanism for keeping the core protocol as minimal as possible. I'll try and flesh that out soon.

alimanfoo:
I'm going to merge this to provide something to build on, but very happy to discuss and revisit any aspect of this as we move forward.

axtimwalde:
I would not include soft and hard links here as their beavior is specific to various backends.  I also find it attractive to think about links on the filesystem as a vehicle to implement non-redundant versioned data but that is a different issue...

docs/protocol/core/v3.0.rst
```diff
@@ -1,3 +1,152 @@
...
+An array is associated with an *encoding pipeline*, which is a
+sequence of zero or more *codecs* that transforms the binary
+representation of a chunk in some way. For example, an encoding
+pipeline might include a checksum codec to ensure data integrity, and
+a compression codec to reduce data size. All codecs implement a common
+*codec interface* which provides a pair of operations, one to perform
+the transformation (encode), the other to reverse the transformation
+(decode).
+
+Each node in a hierarchy is represented by a *metadata document*,
```
axtimwalde: I suggest to add: An empty *metadata document* is equivalent with no *metadata document* and means that there is no meta-data associated with the node.  This is only possible for trivial group nodes.

alimanfoo: Interesting, might need to unpack and discuss a little. 

FWIW in zarr v2 the presence of a metadata document indicates the existence of a node. E.g., if  the key `/foo/bar/.zgroup` exists in the store then that implies a group exists in the hierarchy at logical path `/foo/bar`. Similarly if a key exists in the store at `/foo/bar/baz/.zarray` then that implies an array exists in the hierarchy at logical path `/foo/bar/baz`. So you can actually construct the hierarchy just by knowing which keys are present in the store, without even retrieving or reading the metadata documents. That sounds slightly different from what you're suggesting here?

docs/protocol/core/v3.0.rst
```diff
@@ -1,3 +1,152 @@
...
+
+Conceptual model
+----------------
+
+A Zarr *hierarchy* is a tree structure, where each node in the tree is
+either a *group* or an *array*. Group nodes may have children
+but array nodes may not.
+
+Each node in a hierarchy has a *name* which is a string of ASCII
+characters with some additional constraints. Two sibling nodes cannot
```
joshmoore: Are there any limitations on the characters in that string?

PR #18, opened 2019-05-07
Core protocol v3.0 - data types

alimanfoo:
This PR proposes a section of the v3.0 core protocol specification describing a set of data types for array elements.

Some discussion/decision points, for the following, should they be defined in the core protocol or via a protocol extension:

* Fixed length byte string types (corresponding to types like 'S4' in numpy for an array containing length 4 byte strings)?
* Fixed length unicode type (i.e., corresponding to types like '<U4' in numpy for an array containing length 4 unicode code points)?
* Datetime/timedelta data types (corresponding to e.g. 'M8[ns]' and 'm8[ns]' types in numpy)?
* Structured (i.e., struct-like) data types?
* Variable length data types, such as variable length arrays of a primitive type, or variable length byte or text strings?

### DIFF
--- a/docs/protocol/core/v3.0.rst
+++ b/docs/protocol/core/v3.0.rst
@@ -126,99 +126,100 @@ Core data types
 ~~~~~~~~~~~~~~~
 
 .. list-table:: Data types
-    :header-rows: 1
-    * - Identifier
-      - Numerical type
-      - Size (no. bytes)
-      - Byte order
-    * - `bool`
-      - Boolean
-      - 1
-      - None
-    * - `i1`
-      - signed integer
-      - 1
-      - None
-    * - `<i2`
-      - signed integer
-      - 2
-      - little-endian
-    * - `<i4`
-      - signed integer
-      - 4
-      - little-endian
-    * - `<i8`
-      - signed integer
-      - 8
-      - little-endian
-    * - `>i2`
-      - signed integer
-      - 2
-      - big-endian
-    * - `>i4`
-      - signed integer
-      - 4
-      - big-endian
-    * - `>i8`
-      - signed integer
-      - 8
-      - big-endian
-    * - `u1`
-      - signed integer
-      - 1
-      - None
-    * - `<u2`
-      - unsigned integer
-      - 2
-      - little-endian
-    * - `<u4`
-      - unsigned integer
-      - 4
-      - little-endian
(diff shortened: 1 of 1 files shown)


docs/protocol/core/v3.0.rst
```diff
@@ -101,7 +101,128 @@ TODO define constraints on node names
...
+Each data type is associated with an identifier, which can be used in
+metadata documents to refer to the data type. For the data types
+defined in this protocol, the identifier is a simple ASCII
+string. However, protocol extensions may use any JSON value to
+identify a data type.
+
+Core data types
+~~~~~~~~~~~~~~~
+
+.. list-table:: Data types
```
alimanfoo: This should render as a table when built via sphinx.

I know it's a bit repetitive to enumerate every data type explicitly, given that many have both a big- and little-endian form. However for now I thought it simplest to just enumerate them.

alimanfoo:
A straw man for discussion. Note that tentatively I'm suggesting the core protocol is limited to boolean, integer, float and complex types, but that protocol extensions can define other data types. 

So, e.g., this would mean that other features like datetime/timedelta data types, structured (struct) data types, and variable length data types could each be addressed via separate protocol extensions. The idea behind dividing it up this way is just to keep the core protocol small and simple to implement.

docs/protocol/core/v3.0.rst
```diff
@@ -101,7 +101,128 @@ TODO define constraints on node names
...
+    :header-rows: 1
+    * - Identifier
+      - Numerical type
+      - Size (no. bytes)
+      - Byte order
+    * - `bool`
+      - Boolean
```
shoyer: It would be good to clarify the binary encoding used for booleans, e.g., `False` -> `\x00` and `True` -> `\x01`

alimanfoo:
Also realising that I haven't mentioned a fixed length bytes type (i.e., corresponding to types like 'S4' in numpy for an array containing length 4 byte strings) or a fixed length unicode type (i.e., corresponding to types like '<U4' in numpy for an array containing length 4 unicode code points). Up for discussion whether these should be in the core protocol spec.

docs/protocol/core/v3.0.rst
alimanfoo: Done.

alimanfoo:

PR #22, opened 2019-05-07
Core protocol v3.0 - chunk grids

alimanfoo:
This PR adds a section introducing chunk grids and defining regular chunk grids.

### DIFF
--- a/docs/protocol/core/v3.0.rst
+++ b/docs/protocol/core/v3.0.rst
@@ -103,11 +103,72 @@ Data types
 
 TODO define core data types
 
-Regular chunk grids
--------------------
+Chunk grids
+-----------
 
-TODO define regular chunk grids, including how to form a key for each chunk in a grid
+A chunk grid defines a set of chunks which contain the elements of an
+array. The chunks of a grid form a tessellation of the array space,
+which is a space defined by the dimensionality and shape of the
+array. This means that every element of the array is a member of one
+chunk, and there are no gaps or overlaps between chunks.
 
+In general there are different possible types of grids. The core
+protocol defines the regular grid type, where all chunks are
+hyperrectangles of the same shape. Protocol extensions may define
+other grid types, such as rectilinear grids where chunks are still
+hyperrectangles but do not all share the same shape.
+
+A grid type also defines rules for constructing a unique key for each
+chunk, which is a string of ASCII characters that can be used to save
+and retrieve chunk data in a store.
+
+Regular grids
+~~~~~~~~~~~~~
+
+A regular grid is a type of grid where an array is divided into chunks
+such that each chunk is a hyperrectangle of the same shape. The
+dimensionality of the grid is the same as the dimensionality of the
+array. Each chunk in the grid can be addressed by a tuple of positive
+integers (i, j, k, ...) corresponding to the indices of the chunk
+along each dimension.
+
+The origin vertex of a chunk has coordinates in the array space (i *
+dx, j * dy, k * dz, ...) where (dx, dy, dz, ...) are the grid spacings
+along each dimension, also known as the chunk shape. Thus the origin
+vertex of the chunk at grid index (0, 0, 0, ...) is at coordinate (0,
+0, 0, ...) in the array space, i.e., the grid is aligned with the
+origin of the array. If the length of any array dimension is not
+perfectly divisible by the chunk length along the same dimension, then
+the grid will overhang the edge of the array space.
+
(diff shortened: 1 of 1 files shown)


alimanfoo:
Straw man for discussion. Note that I'm tentatively suggesting that the core protocol sticks to defining regular chunk grids, however protocol extensions can define other grid types. For example, rectilinear grids where chunks may have different shapes would be addressed via a protocol extension. Similarly, grids where the origin of the array may occur anywhere in any chunk (needed to allow arrays to "grow" in the "negative" direction) would be a protocol extension. Again, just thinking of ways to keep the core protocol simple and minimal but allow for flexibility and development of additional features via protocol extensions.

docs/protocol/core/v3.0.rst
```diff
@@ -103,11 +103,72 @@ Data types
...
-Regular chunk grids
--------------------
-
-TODO define regular chunk grids, including how to form a key for each chunk in a grid
-
+Chunk grids
+-----------
+
+A chunk grid defines a set of chunks which contain the elements of an
+array. The chunks of a grid form a tessellation of the array space,
```
joshmoore: Do you foresee a _simpler_ grid definition than doesn't include the relationship to an array?

docs/protocol/core/v3.0.rst
```diff
@@ -103,11 +103,72 @@ Data types
...
+chunk, and there are no gaps or overlaps between chunks.
+
+In general there are different possible types of grids. The core
+protocol defines the regular grid type, where all chunks are
+hyperrectangles of the same shape. Protocol extensions may define
+other grid types, such as rectilinear grids where chunks are still
+hyperrectangles but do not all share the same shape.
+
+A grid type also defines rules for constructing a unique key for each
+chunk, which is a string of ASCII characters that can be used to save
```
joshmoore: This is perhaps for a different PR, but I could see specifying a "tuple of strings of ASCII characters in the set [...]". See below.

docs/protocol/core/v3.0.rst
```diff
@@ -103,11 +103,72 @@ Data types
...
+
+An element of an array with coordinates (i, j, k, ...) will occur
+within the chunk at grid index (i // dx, j // dy, k // dz, ...), where
+// is the floor division operator. The element will have coordinates
+(i % dx, j % dy, k % dz, ...) within that chunk. For example, @@TODO
+example.
+
+The key for chunk with grid index (i, j, k, ...) is formed by
+concatenating the ASCII string representation of each index, joined
+together via the period (".") character. For example, in a 3
```
joshmoore: Related to the "tuple of strings" from above, I would propose either leaving the joining operation to the backend or going so far as to suggest "/" as the default. My understanding is that cloud storage doesn't suffer under use of "/" but local storage _does_ suffer under use of ".".

alimanfoo: I've just tentatively pushed an edit which allows for an array to have configurable prefix, suffix and separator for chunk keys, which would be a mechanism to allow e.g. use of "/" as chunk key separator. Happy to row back on this if anyone feels that should be a protocol extension.

docs/protocol/core/v3.0.rst
alimanfoo: Ah OK. IIUC what @axtimwalde was saying was that there are some useful functionalities that could be provided without needing to know anything about the grid layout. For example, if you wanted to copy data from one store to another, or recode chunks using a different compressor. But I think the core protocol needs to define the full picture, in the sense that chunks always belong to an array, and that array will have some grid layout that defines what the chunks contain.

docs/protocol/core/v3.0.rst
```diff
@@ -103,11 +103,81 @@ Data types
...
+indices and suffix are joined using a separator. The default value for
+the prefix is the empty string (""), the default value for the
+separator is the period character (".") and the default value for the
+suffix is the empty string (""), but these values may be configured,
+see the section on `Array metadata`_ below.
+
+For example, in a 3 dimensional array at path "/foo/bar" configured
+with default values for the chunk key prefix, suffix and separator,
+the key for the chunk at grid index (1, 23, 45) is the string
+"/foo/bar/1.23.45".
```
alimanfoo: @axtimwalde, @constantinpape, @funkey, do you want to add a configuration option to allow the chunk indices to be given in reverse order, as in n5? Or is it OK to fix on using a single ordering as in zarr v2?

alimanfoo:

PR #24, opened 2019-05-08
Core protocol v3.0 - chunk memory layouts

alimanfoo:
This PR adds draft content for the core protocol v3.0 section on memory layouts.

### DIFF
--- a/docs/protocol/core/v3.0.rst
+++ b/docs/protocol/core/v3.0.rst
@@ -145,9 +145,9 @@ F contiguous memory layout
 
 In this memory layout, the binary values of the array elements are
 organised into a sequence such that the first dimension of the array
-is the fastest changing dimension, also known as "column-major" or
-"Fortran" order. This layout is only applicable to arrays with fixed
-size data types.
+is the fastest changing dimension, also known as "column-major"
+order. This layout is only applicable to arrays with fixed size data
+types.
 
 For example, for a two-dimensional array with chunk shape (dx, dy),
 the binary values for a given chunk are taken from chunk elements in
--- a/docs/protocol/core/v3.0.rst
+++ b/docs/protocol/core/v3.0.rst
@@ -112,7 +112,48 @@ TODO define regular chunk grids, including how to form a key for each chunk in a
 Memory layouts
 --------------
 
-TODO define "C" and "F" memory layouts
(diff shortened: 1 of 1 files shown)


docs/protocol/core/v3.0.rst
```diff
@@ -112,7 +112,48 @@ TODO define regular chunk grids, including how to form a key for each chunk in a
...
+
+In this memory layout, the binary values of the array elements are
+organised into a sequence such that the last dimension of the array is
+the fastest changing dimension, also known as "row-major" order. This
+layout is only applicable to arrays with fixed size data types.
+
+For example, for a two-dimensional array with chunk shape (dx, dy),
+the binary values for a given chunk are taken from chunk elements in
+the order (0, 0), (0, 1), (0, 2), ..., (dx - 1, dy - 3), (dx - 1, dy -
+2), (dx - 1, dy - 1).
```
martindurant: Maybe worth mentioning here, that the memory offset/location within the chunk is determined by this ordering *and* the size of the dtype. Will endianness be in the dtype section only (I know it's mentioned already). That could be be loosely considered memory layout, especially for sub-byte types.

docs/protocol/core/v3.0.rst
```diff
@@ -112,7 +112,48 @@ TODO define regular chunk grids, including how to form a key for each chunk in a
...
+
+F contiguous memory layout
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In this memory layout, the binary values of the array elements are
+organised into a sequence such that the first dimension of the array
+is the fastest changing dimension, also known as "column-major" or
+"Fortran" order. This layout is only applicable to arrays with fixed
```
martindurant: It is made specific what "F" stands for here. Either could add that "C" comes from the programming language of that name, or simply not mention Fortran at all.

alimanfoo:

docs/protocol/core/v3.0.rst
alimanfoo: Sounds good. Could you suggest an edit?

docs/protocol/core/v3.0.rst
alimanfoo: Fair point, have removed "Fortran".

PR #25, opened 2019-05-09
Core protocol v3.0 - node names

alimanfoo:
This PR adds a draft for the section on node names.

### DIFF
--- a/docs/protocol/core/v3.0.rst
+++ b/docs/protocol/core/v3.0.rst
@@ -10,8 +10,9 @@ either a *group* or an *array*. Group nodes may have children
 but array nodes may not.
 
 Each node in a hierarchy has a *name* which is a string of ASCII
-characters with some additional constraints. Two sibling nodes cannot 
-have the same name. The root node does not have a name.
+characters with some additional constraints defined in the section on
+`Node names`_ below. Two sibling nodes cannot have the same name. The
+root node does not have a name.
 
 Each node in a hierarchy has a *path* which uniquely identifies that
 node and defines its location within the hierarchy. The path is formed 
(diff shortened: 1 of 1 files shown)


docs/protocol/core/v3.0.rst
```diff
@@ -95,7 +96,27 @@ provide.
...
-TODO define constraints on node names
+Except for the root node, each node in a hierarchy must have a name,
+which is a string of ASCII characters. To ensure consistent behaviour
+across different storage systems, the following constraints apply to
+node names:
+
+* must not be the empty string ("")
+
+* must consist only of characters in the sets `a-z`, `A-Z`, `0-9`,
+  `-_.`
```
alimanfoo: Is this too restrictive? I was thinking that node names are used in file paths when using file system storage, and so we should ensure no characters are used that could be problematic within a file system path. But are there other characters that people use in dataset naming that would be safe to include?

docs/protocol/core/v3.0.rst
```diff
@@ -95,7 +96,27 @@ provide.
...
+  ".."
+
+* must be at most 255 characters long
+
+Note that node names are used to form storage keys, and that some
+storage systems will perform a case-insensitive comparison of storage
+keys during retrieval. Therefore, within a hierarchy, all nodes within
+a set of sibling nodes must have a name that is unique under
+case-insensitive comparison. E.g., the names "foo" and "FOO" are not
+allowed for sibling nodes.
```
alimanfoo: This restriction is to deal with the situation where file system storage is used on a platform with a case insensitive file system, e.g., current default on mac I believe.

docs/protocol/core/v3.0.rst
```diff
@@ -95,7 +96,27 @@ provide.
...
+across different storage systems, the following constraints apply to
+node names:
+
+* must not be the empty string ("")
+
+* must consist only of characters in the sets `a-z`, `A-Z`, `0-9`,
+  `-_.`
+
+* must not be a string composed only of period characters, e.g. "." or
+  ".."
```
alimanfoo: Again this is to avoid node names that would be problematic in a file path.

docs/protocol/core/v3.0.rst
```diff
@@ -95,7 +96,27 @@ provide.
...
+
+* must be at most 255 characters long
```
alimanfoo: Again this is to avoid node names that would be problematic in a file path.

alimanfoo:

PR #28, opened 2019-05-11
Core protocol v3.0 - codecs

alimanfoo:
This PR adds a straw man draft section on codecs.


alimanfoo:

PR #29, opened 2019-05-14
Core protocol v3.0 - array and group metadata

alimanfoo:
This PR has a straw man for a section on array and group metadata.

### DIFF
--- a/docs/protocol/core/v3.0.rst
+++ b/docs/protocol/core/v3.0.rst
@@ -123,7 +123,60 @@ TODO define the codec interface
 Array metadata
 --------------
 
-TODO define the structure and content of array metadata documents
+An array metadata document is a JSON document containing configuration
+metadata for an array. The document must contain a single object, here
+called the metadata object. The metadata object must contain the
+following keys:
+
+`zarr_format`
+
+    Version number of the Zarr core protocol that defines the metadata
+    format. For arrays conforming to this specification, the value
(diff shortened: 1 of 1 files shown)


alimanfoo:
This is still WIP but thought I'd surface to give a sense of direction of travel. Comments welcome.

alimanfoo:
Did a bit of refactoring and added in a subsection about group metadata.

alimanfoo:
One change I've tentatively pushed is to allow for inclusion of "user attributes" within the array and group metadata documents, a la n5, rather than in separate documents a la zarr v2.

alimanfoo:

joshmoore:
No objections, but I didn't work my way through this one as closely as I should have.

tam203:
If I read this correctly the user/application attributes are no longer at a separate/object from the Zarr required ones. If this is indeed the case then I think that's a great step.

Good stuff @alimanfoo

alimanfoo:
> No objections, but I didn't work my way through this one as closely as I should have.

Thanks, no problem, I think we can revisit this as many times as we want after merge to the 3.0 dev branch.

alimanfoo:
> If I read this correctly the user/application attributes are no longer at a separate/object from the Zarr required ones. If this is indeed the case then I think that's a great step.

Cool, yes proposing that attributes are within the array and group metadata objects, not separate.

PR #31, opened 2019-05-21
Core protocol v3.0 - add metadata examples

alimanfoo:
Adds array and group metadata examples.


alimanfoo:
Going to merge this one in but very happy to discuss and revisit in the dev branch.

PR #32, opened 2019-05-21
Core protocol v3.0 - simplify chunk grids

alimanfoo:
This PR simplifies the chunk grids section a little. In particular, the section now no longer provides the rules for defining the full chunk store key, leaving that for the later storage protocol section, but now just defines how to create a chunk identifier like "1.23.45" which can be used as the last segment of a key.


alimanfoo:
Going to merge this in to have a consistent base for discussion, but very happy to review and revise after merge.

PR #34, opened 2019-05-23
Core protocol v3.0 - front matter tweaks

alimanfoo:
This PR simplifies the front matter slightly, and adds a placeholder for a change log section at the end.


alimanfoo:
Hopefully uncontroversial, merging now but happy to revisit if any comments or suggestions.

PR #35, opened 2019-05-23
Core protocol v3.0 - illustrate data type extension (datetimes and timedeltas)

alimanfoo:
This PR is intended to provide an illustration of how protocol extensions that define new data types would be documented and used together with the v3.0 protocol.

This PR uses the example of datatime and timedelta data types. This is meant for illustration only at this stage.


alimanfoo:
Viewable here: https://zarr-specs.readthedocs.io/en/core-protocol-v3.0-illustrate-dtype-extensions/index.html

alimanfoo:
I'm going to merge this into the 3.0 dev branch just to have content together for review. Very happy to split apart again later if it makes more sense to do so.

PR #47, opened 2019-07-26
File system store spec

alimanfoo:
This PR adds a spec for a file system store. This is intended as an illustration of what a store spec could look like.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,6 @@
 
 # sphinx build files
 docs/_build
+
+# pycharm
+.idea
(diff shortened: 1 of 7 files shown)


alimanfoo:
I'm thinking it will be useful to have an example of a storage spec to point to, to illustrate the idea of having a separate spec for each possible storage implementation (file system, S3, GCS, ABS, LMBD, etc.) and how this fits alongside the core protocol v3.0 spec. I will merge this shortly into the core-protocol-v3.0-dev branch to have it built on RTFD as something we can point to, but this ultimately should get moved into a separate branch and reviewed properly. Comments very welcome at any time, and we will revisit this before it goes anywhere near master branch.

PR #52, opened 2020-03-27
Use pydata-sphinx-theme

jrbourbeau:
This PR switches to using the [`pydata-sphinx-theme`](https://github.com/pandas-dev/pydata-sphinx-theme) Sphinx theme. This is totally subjective, but I do like the way this theme looks (see screenshot below). Curious to hear what others think. 

Screen Shot 2020-03-26 at 9 14 53 PM


jakirkham:
Nice! I'm convinced 😄

cc @ryan-williams @joshmoore @alimanfoo (thoughts? 🙂)

joshmoore:
vs. 
Screen Shot 2020-03-27 at 09 13 11

? Yeah, I'm in :)

As a side note, I'm willing to look into getting this branch to be the default view on readthedocs.

alimanfoo:
Big :+1: :-)

jrbourbeau:
Thanks for the feedback! Merging

> I'm willing to look into getting this branch to be the default view on readthedocs.

That'd be great. If you'd like, feel free to add me on readthedocs (username jrbourbeau) and I can also take a look

joshmoore:
If you omit `latest` from RTD, it now redirects to https://zarr-specs.readthedocs.io/en/core-protocol-v3.0-dev/ to make it explicit. We may want to move `stable` to point at `master` and have `latest` point at `v3.0` or `dev` or whatever we will be using going down the road.

alimanfoo:
> If you omit `latest` from RTD, it now redirects to https://zarr-specs.readthedocs.io/en/core-protocol-v3.0-dev/ to make it explicit. 

Thanks Josh.

> We may want to move `stable` to point at `master` and have `latest` point at `v3.0` or `dev` or whatever we will be using going down the road.

Yep, not sure what is the best approach until we have something stable.

jakirkham:
Maybe we can disable `stable` from showing up or make `latest` the default or similar?

alimanfoo:
> Maybe we can disable `stable` from showing up or make `latest` the default or similar?

Yeah, maybe `core-protocol-v3.0-dev` should be the default branch for the time being?

alimanfoo:
...or disable `latest` altogether?

joshmoore:
> Maybe we can disable stable from showing up or make latest the default or similar?

`latest` was the default this morning. Trying v3 now as the default.

> ...or disable latest altogether?

I deactivated `latest`, but it's still reachable, presumably because there was already a build?

alimanfoo:
All looks good now, thanks Josh.

PR #54, opened 2020-03-30
More updates to core protocol document

jrbourbeau:
This PR contains another batch of updates from reading through the spec and implments some suggestions from review comments in https://github.com/zarr-developers/zarr-specs/pull/16. Specifically: 

- Use BCE/CE instead of BC/AD in datetime units table (https://github.com/zarr-developers/zarr-specs/pull/16#discussion_r297878854)

- Remove complex numbers from protocol data types in text (was already removed from core data types table)

- Updates the term "bootstrap metadata" to "entry point metadata" as suggested in https://github.com/zarr-developers/zarr-specs/pull/16#issuecomment-501027925. We can go with something else here too, I just happen to like "entry point" more than "bootstrap"

- Removes reference to ASCII in node names section as suggested in https://github.com/zarr-developers/zarr-specs/pull/16#discussion_r286450533

- Small typos

- `http://` -> `https://`

Happy to remove any changes here is they merit further discussion on the v3.0 PR


docs/protocol/core/v3.0.rst
```diff
@@ -257,7 +257,7 @@ Node names
 ==========
 
 Except for the root node, each node in a hierarchy must have a name,
-which is a string of ASCII characters. To ensure consistent behaviour
+which is a string characters. To ensure consistent behaviour
```
alimanfoo: Maybe we need to say that it is a string of Unicode characters? I.e., we don't say anything about encoding, but we do say what the character set is?

docs/protocol/core/v3.0.rst
```diff
@@ -575,33 +574,31 @@ given in the section below on `Array metadata`_.
 Metadata
 ========
 
-This section defines the structure of metadata documents for Zarr
-arrays and groups. Metadata documents are defined here using the JSON
+This section defines the structure of metadata documents for Zarr hierarchies,
+which consists of three types metadata documents: a entry point metadata document,
```
alimanfoo:
```suggestion
which consists of three types metadata documents: an entry point metadata document,
```

docs/protocol/core/v3.0.rst
```diff
@@ -575,33 +574,31 @@ given in the section below on `Array metadata`_.
...
-Bootstrap metadata
-------------------
 
-@@TODO should this be called "entry-point metadata"? Or something
-else?
+Entry point metadata
+--------------------
 
-Each Zarr hierarchy must have a bootstrap metadata document, which
+Each Zarr hierarchy must have a entry point metadata document, which
```
alimanfoo:
```suggestion
Each Zarr hierarchy must have an entry point metadata document, which
```

docs/protocol/core/v3.0.rst
```diff
@@ -653,7 +650,7 @@ containing the following names:
     ``must_understand`` is ``true`` then processing must terminate and
     an appropriate error raised.
 
-For example, below is a bootstrap metadata document, specifying that
+For example, below is a entry point metadata document, specifying that
```
alimanfoo:
```suggestion
For example, below is an entry point metadata document, specifying that
```

docs/protocol/core/v3.0.rst
```diff
@@ -662,7 +659,7 @@ JSON is being used for encoding of group and array metadata::
         "extensions": []
     }
 
-For example, below is a bootstrap metadata document as above, but also
+For example, below is a entry point metadata document as above, but also
```
alimanfoo:
```suggestion
For example, below is an entry point metadata document as above, but also
```

docs/protocol/core/v3.0.rst
```diff
@@ -808,7 +807,7 @@ compressed using gzip compression prior to storage::
         },
         "chunk_memory_layout": "C",
         "compressor": {
-            "codec": "http://purl.org/zarr/spec/codec/gzip",
+            "codec": "https://purl.org/zarr/spec/codec/gzip",
```
alimanfoo: Update to include version number in URI, as per codec spec draft?

docs/protocol/core/v3.0.rst
alimanfoo: Btw happy to merge this PR and return to this. I think there is a broader discussion needed around character sets and encodings.

docs/protocol/core/v3.0.rst
jrbourbeau: Sounds good, I'll include that in the gzip codec PR

docs/protocol/core/v3.0.rst
jrbourbeau: There's an existing character set in a bullet point a few lines below:

> must consist only of characters in the sets `a-z`, `A-Z`, `0-9`, `-_.`

I agree we should return to this section after discussing character sets and encodings more broadly

jrbourbeau:
Thanks for reviewing @alimanfoo!

alimanfoo:
Thanks @jrbourbeau, great to be moving on this again.

PR #55, opened 2020-03-30
Add initial draft for gzip codec

jrbourbeau:
This PR adds a codec spec for Gzip. Any feedback is much appreciated.


docs/codecs/gzip/v1.0.rst
```diff
@@ -0,0 +1,120 @@
...
+-----------------------------
+
+Specification URI:
+    https://purl.org/zarr/spec/codecs/gzip/1.0
+Issue tracking:
+    `GitHub issues <https://github.com/zarr-developers/zarr-specs/labels/codecs-gzip-v1.0>`_
+Suggest an edit for this spec:
+    `GitHub editor <https://github.com/zarr-developers/zarr-specs/blob/core-protocol-v3.0-dev/docs/codecs/gzip/v1.0.rst>`_
+
+Copyright 2019 `Zarr core development
```
alimanfoo:
```suggestion
Copyright 2020 `Zarr core development
```

docs/codecs/gzip/v1.0.rst
```diff
@@ -0,0 +1,120 @@
+========================
+Gzip Codec (version 1.0)
+========================
+-----------------------------
+ Editor's draft 26 July 2019
```
alimanfoo:
```suggestion
 Editor's draft 31 March 2020
```

docs/codecs/gzip/v1.0.rst
```diff
@@ -0,0 +1,120 @@
...
+References
+==========
+
+.. [RFC2119] S. Bradner. Key words for use in RFCs to Indicate
+   Requirement Levels. March 1997. Best Current Practice. URL:
+   https://tools.ietf.org/html/rfc2119
+
+.. [RFC1951] P. Deutsch. DEFLATE Compressed Data Format Specification version
+   1.3. Requirement Levels. May 1996. Informational. URL:
+   https://tools.ietf.org/html/rfc1952
```
alimanfoo:
```suggestion
   https://tools.ietf.org/html/rfc1951
```

docs/codecs/gzip/v1.0.rst
```diff
@@ -0,0 +1,120 @@
...
+this specification are introduced with the words "for example".
+
+
+Chunk encoding/decoding with Gzip
+=================================
+
+@@TODO define how chunks are encoded and decoded
+
+Chunks are encoded and decoded using the Gzip compression algorithm defined in
+[RFC1951]_ and [RFC1952]_. The compression level is an integer from 0 to 9
```
alimanfoo: I don't fully understand all the details here, but this spec will need to be clear I think about the fact that the encoded data should conform to the gzip file format, i.e., include the header and trailer, and not be just a DEFLATE data stream. 

To put it another way, currently in numcodecs we have both a zlib and a gzip codec, which are different. (It was my error originally to think these could be the same.) This spec should match the behaviour of the numcodecs gzip codec.

alimanfoo: FWIW I think it would be OK to merge this PR without fully working out how to pin down exactly all of the details. But I think it would be worth adding some clarifying statement here just to make sure we have some sense of the distinction.

docs/codecs/gzip/v1.0.rst
```diff
@@ -0,0 +1,120 @@
...
+
+Additionally, the compression level must be specified as the value of the
+``level`` name in the ``configuration`` metadata name. For example, the array
+metadata document below specifies a Gzip codec configured with a compression
+level of 1::
+
+
+    {
+        "compressor": {
+            "codec": "http://purl.org/zarr/spec/codecs/gzip/1.0",
```
alimanfoo: Probably worth updating the example in the core protocol spec as well to match this and use the full URI.

jrbourbeau: Done

docs/codecs/gzip/v1.0.rst
jrbourbeau: Thanks for the feedback on this @alimanfoo. I've updated the text to be a little more clear about conforming to the gzip file format and added another `@@TODO` for that specific point

jrbourbeau:
Thanks for reviewing @alimanfoo! I've applied your code review suggestions and have added an additional `@@TODO` for the point your raised about conforming to the gzip file format

alimanfoo:
Thanks @jrbourbeau :+1:

Issue #56, opened 2020-03-31
Node name character set


alimanfoo:
Currently the core protocol v3.0 draft includes a [section on node names](https://zarr-specs.readthedocs.io/en/core-protocol-v3.0-dev/protocol/core/v3.0.html#node-names) which defines the set of characters that may be used to construct the name of a node (array or group).

The set of allowed characters is currently extremely narrow, being only the union of `a-z`, `A-Z`, `0-9`, `-_.`. There is no support for non-latin characters, which obviously creates a barrier for many people. I'd like to relax this and allow any Unicode characters. Could we do this, and if we did, what problems would we need to anticipate and address in the core protocol spec? 

Some points to bear in mind for this discussion:

Currently in the core protocol, node names (e.g., "bar") are used to form node paths (e.g., "/foo/bar"), which are then used to form storage keys for metadata documents (e.g., "meta/root/foo/bar.array") and data chunks (e.g., "data/foo/bar/0.0"). These storage keys may then be handled by a variety of different store types, including [file system stores](https://zarr-specs.readthedocs.io/en/core-protocol-v3.0-dev/stores/filesystem/v1.0.html) where storage keys are translated into file paths, cloud object stores where storage keys are translated into object keys, etc.

Different store types will have different abilities to support the full Unicode character set for node names. For example, although most file systems support Unicode file names, there are still [reserved characters and words](https://en.wikipedia.org/wiki/Filename#Reserved_characters_and_words) which cannot be used, which differ between operating systems and file system types. However, these constraints might not apply at all to other store types, such as cloud object stores. In general, other store types may have different constraints. Do we need to anticipate any of this, or can we delegate these issues to be dealt with in the different store specs?

One last thought, whatever we decide, the set of allowed characters should probably be defined with respect to some standard character set, e.g., Unicode. I.e., we should probably reference the appropriate standard when discussing which characters are allowed.

Carreau:
> Could we do this, and if we did, what problems would we need to anticipate and address in the core protocol spec?

The main problem you will hit in my opinion is Unicode normalisation, would `é` be `é` or `e`+ `´` ? I'm also worried about unicode support for some edge languages like fortran.

[...]

PR #60, opened 2020-04-26
Fix rendering of allowed characters.

Carreau:
The default rule for sphinx with simple backticks seem to be "cite",
which is rendered in italic, making it relatively hard to distinguish
what is the set of characters meant vs, formatting.

On my initial read on current page I actually though the trailing dot
was finishing the sentence, and it was not clear whether or not spaces
(and or commas) were part of allowed character.

With the double quote each character group is rendered: monospace, and in
a different color (for current theme)

--- 

Sidenote, I see no mention of encoding in this section, I'm guessing this character set has been chosen to be "safe" in the sens ascii-compatible right ? In which case you might want to actually specify the bytes range maybe as not all encoding will get these as the same bytes:

```
In [1]: from string import ascii_letters, digits
In [2]: (ascii_letters + digits).encode().decode('utf16')
Out[2]: '扡摣晥桧橩汫湭灯牱瑳癵硷穹䉁䑃䙅䡇䩉䱋乍偏剑呓噕塗婙\u3130㌲㔴㜶㤸'
```

### DIFF
--- a/docs/protocol/core/v3.0.rst
+++ b/docs/protocol/core/v3.0.rst
@@ -263,8 +263,8 @@ node names:
 
 * must not be the empty string ("")
 
-* must consist only of characters in the sets `a-z`, `A-Z`, `0-9`,
-  `-_.`
+* must consist only of characters in the sets ``a-z``, ``A-Z``, ``0-9``,
+  ``-_.``
 
 * must not be a string composed only of period characters, e.g. "." or
   ".."


joshmoore:
Thanks, @Carreau. Recently ran into the same issue myself. Merging proactively, and I'll paste a screenshot from readthedocs.org as a _post hoc_ review.

joshmoore:
Screen Shot 2020-04-27 at 11 40 51

alimanfoo:
Many thanks @Carreau.

> Sidenote, I see no mention of encoding in this section, I'm guessing this character set has been chosen to be "safe" in the sens ascii-compatible right ? In which case you might want to actually specify the bytes range maybe as not all encoding will get these as the same bytes

This is something that needs some clarification. I think encoding is an issue for the store implementation, i.e., the core protocol does not need to specify an encoding, just a set of characters that are allowed. But that needs explanation.

(Also I think the set of allowed characters should be expanded, xref #56.)

PR #69, opened 2020-05-14
Reiterate some of the examples in table form.

Carreau:
It seem there was a typo/error where ocreating a key was suggesting to
concatenate "meta" instead of meta/root if my reading is correct.

-- 

Questions while I'm there

1) why the different logic for root and have `meta/root.group` instead of `meta/root/.goup` avoid leading dots? 

2) the path "P" are between single backticks, which representation depends on sphinx default role. Was the original meaning double ticks for code or italic ? 

3) The implicit groups should likely be refined. If you create a dataset with implicit groups it is not clear if the store can, can not, should, should not, must, or must not create the intermediate `.groups` keys

For me the following implies that all groups have a valid metadata document:

```
To create a group at hierarchy path P, perform set(group_meta_key(P), value), where value is the serialisation of a valid group metadata document.
```

Which is contradictory with I believe the implicit groups. 

It may be that in the above you mean users can create an _empty_ group that will persists, and that if that document is not there, then the group may be automatically removed on dataset removal – which seem natural for a flat-ish storage of keys; though it should likely be clarified what store are allowed or must do on deletions.

### DIFF
--- a/docs/protocol/core/v3.0.rst
+++ b/docs/protocol/core/v3.0.rst
@@ -440,6 +440,18 @@ grid will be (2, 10, 8), meaning that there will be 2 chunks along the
 first dimension, 10 along the second dimension, and 8 along the third
 dimension.
 
+.. list-table:: Regular Grid Example
+    :header-rows: 1
+
+    * - Array Shape
+      - Chunk Shape
+      - Chunk Grid Shape
+      - Notes
+    * - (10, 200, 3000)
+      - (5, 20, 400)
+      - (2, 10, 8)
+      - The grid does overhang the edge of the array on the 3rd dimension.
+
 An element of an array with coordinates (`a`, `b`, `c`, ...) will
 occur within the chunk at grid index (`a` // `dx`, `b` // `dy`, `c` //
 `dz`, ...), where "//" is the floor division operator. The element
@@ -449,6 +461,7 @@ that chunk, where "%" is the modulo operator. For example, if a
 (5, 20, 400), then the element of the array with coordinates (7, 150, 900)
 is contained within the chunk at grid index (1, 7, 2) and has coordinates
 (2, 10, 100) within that chunk.
+
 
 The identifier for chunk with grid index (`i`, `j`, `k`, ...) is
 formed by joining together ASCII string representations of each index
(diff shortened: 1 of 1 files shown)


alimanfoo:
Brief response on this:

> 1. why the different logic for root and have `meta/root.group` instead of `meta/root/.goup` avoid leading dots?

Main reason is, how can you discover the children of the root group?

With the current key design, `list_dir("/meta/root/")` will give the the result. 

More generally, how can you discover the children of an arbitrary group at hierarchy path P?

Perform `list_dir("meta/root" + P + "/")`.

Happy to expand on this, but basic idea is that key design allows top-down exploration of a hierarchy via a minimum number of calls to the store API, assuming that the store implements the `list_dir()` operation. 

Storage key for the root group metadata document gets a special case because the root group has no name.

alimanfoo:
Worked example [here](https://zarr-specs.readthedocs.io/en/core-protocol-v3.0-dev/protocol/core/v3.0.html#protocol-operations). Copying here for convenience:

> To discover the children of a group at hierarchy path `P`, perform
> ``list_dir("meta/root" + P + "/")``. Any returned key ending in
> ".array" indicates an array. Any returned key ending in
> ".group" indicates a group. Any returned prefix indicates a
> child group implied by some descendant.
> 
> For example, if a group is created at path "/foo/bar" and an array
> is created at path "/foo/baz/qux", then the store will contain the
> keys "meta/root/foo/bar.group" and "meta/root/foo/bar/baz/qux.array". Groups
> at paths "/", "/foo" and "/foo/baz" have not been explicitly
> created but are implied by their descendants. To list the children
> of the group at path "/foo", perform ``list_dir("meta/root/foo/")``,
> which will return the key "meta/root/foo/bar.group" and the prefix
> "meta/root/foo/baz/". From this it can be inferred that child groups
> "/foo/bar" and "/foo/baz" are present.

alimanfoo:
> 2\. the path "P" are between single backticks, which representation depends on sphinx default role. Was the original meaning double ticks for code or italic ?

Not sure, I think my style got a bit confused, happy to be consistified.

alimanfoo:
> 3\. The implicit groups should likely be refined. If you create a dataset with implicit groups it is not clear if the store can, can not, should, should not, must, or must not create the intermediate `.groups` keys

Yes would be good to clarify this. My intention was that an implementation may choose to explicitly create all ancestor groups, but that is not necessary. 

Just to clarify language, I use "explicit group" to mean a group for which a group metadata document exists in the store, and "implicit group" to mean a group implied by some descendant which is an explicit group or array.

> For me the following implies that all groups have a valid metadata document:
> 
> To create a group at hierarchy path P, perform set(group_meta_key(P), value), where value is the serialisation of a valid group metadata document.
>
> Which is contradictory with I believe the implicit groups.

Maybe that should read, "to create an explicit group at hierarchy path P, ...".

PR #71, opened 2020-05-20
Suggestion changes to the spec,

Carreau:
This is rough suggestions, I am happy to change any naming/section or
remove of course.

This is not particularly organised, but intended to get the discussion
going. Feel free to push chaanges to this branch

### DIFF
--- a/docs/protocol/core/v3.0.rst
+++ b/docs/protocol/core/v3.0.rst
@@ -72,6 +72,11 @@ Unlike Zarr spec version 2, the spec V3 have mainly the following difference:
   - V3 has an explicit root, while v2 roots and group could not be distinguished. 
   - Separation of the data and  metadata key space. 
   - Explicit support for extensions.
+  - chunk separator is ``/`` by default.
+
+This means that a store cannot be opened at an arbitrary point, and need to be
+opened at the root. User facing convenience functions could walk up a given
+hierarchy and return a sub-group, but this is not part of the API.
 
 
 Document conventions
@@ -146,6 +151,11 @@ conceptual model underpinning the Zarr protocol.
     "/foo/bar" identifies a node named "bar", whose parent is named
     "foo", whose parent is the root of the hierarchy. The path "/"
     identifies the root node.
+    
+    A path always stars with ``/``.
+
+    QUESTION: do we want to codify that group ``path`` end, can end, or must end
+    in ``/`` ?
 
 .. _dimension:
 .. _dimensions:
@@ -424,14 +434,14 @@ defined by IEEE 754-2008.
 Additionally to these base types, an implementation should also handle the
 raw/opaque pass through type designated by the lowercase letter ``r`` followed
 by the number of bits. For example, ``r3``, ``r11`` should be understood as
-fallback types of respectively 3 and 11 bytes longs, and implementation should
+fallback types of respectively 3 and 11 bits longs, and implementation should
 ensure to pass or accept those types unchanged when using extension fallback.
-For array whose total length is not a multiple of 8, zarr implementation can
+For array whose total length is not a multiple of 8, Zarr implementation can
 expect expects a buffer which length is a multiple of 8bits, padded with
 appropriate value at the end, and will in the same way return buffers padded to
-the next multiple of 8 bytes. 
+the next multiple of 8 bits. 
 
-For example, here is a array of 5 elements of opaque type r3 will be stored as
+For example, here is a array of 5 elements of opaque type ``r3`` will be stored as
 
  -  ``AAABBBCC CDDDEEE_``, 3*5 = 15 bits, rounded to 2 bytes, with arbitrary
    last bit.
@@ -441,8 +451,6 @@ For example, here is a array of 5 elements of opaque type r3 will be stored as
 
 This is meant to help replacing the packbits filter from numcodecs, as well as
 storing categorical data via extension in Zarr v3
-
-
 
 
 Note: currently only fixed size elements are supported as a core data type.
@@ -1049,7 +1057,7 @@ Metadata encoding
 -----------------
 
 The entry point metadata document must be encoded as JSON. The array (``*.array``s) and
-group metadata documents (``*.group``s) must be encoded as per the MIME type given in
+group metadata documents (``*.group``s) must be encoded as per the type given in
 the ``metadata_encoding`` field in the entry point metadata document
 (described below).
 
@@ -1085,9 +1093,12 @@ A store can make the following assumption on the structures of the keys it will 
 
 - A key always:
   - start with ``meta/``
-  - starts  with ``data/``
+  - start  with ``data/``
   - is exactly ``zarr.json``.
 
+- Most of the keys:
+  - start with ``meta/root``
+  - start with ``data/root``
 
 
 - List operations ``list_dir`` will always be passed keys ending with a trailing
@@ -1105,7 +1116,7 @@ For example, a store containing the following keys:
 
 The following queries are invalid:
   - ``list_dir('201')`` is invalid as ``"201"`` is not an existing node.
-  - ``list_dir('2018')`` is invalid querries as ``"2018"`` does not ends with a ``/``,
+  - ``list_dir('2018')`` is invalid queries as ``"2018"`` does not ends with a ``/``,
 
 This is valid:
   - ``list_dir('2018/')``
@@ -1144,7 +1155,9 @@ A **writeable store** supports the following operations:
     | Output: None
 
     
-    clients of delete_prefix should pay attention to pass a trailing slash on the key to delete a complete dataset or group, otherwise the store may delete similar keys.
+    Clients of delete_prefix should pay attention to pass a trailing slash on
+    the key to delete a complete dataset or group, otherwise the store may
+    delete similar keys.
 
 A **listable store** supports any one or more of the following
 operations:
@@ -1185,8 +1198,6 @@ which do not contain the character "/" after the given prefix.
     On non-existing prefix, store may return the empty set.
 
 
-
-
     Note: The requirement on trailing slashes is to avoid 
     search returning keys in the same hierarchy level but longer name, and
     potentially expensive logic testing for the present of trailing slash on
@@ -1196,27 +1207,27 @@ which do not contain the character "/" after the given prefix.
      - /meta/foo/dataset
      - /meta/foobar
 
-     list_dir('foo') == 'foo'&'foobar'
-     list_dir('foo/') == 'foo/dataset'
+     list_dir('/meta/foo') == '/meta/foo'&'/meta/foobar'
+     list_dir('/meta/foo/') == '/meta/foo/dataset'
 
 
     Stores Must return trailing slashes in key responses when those
-    have prefix of other keys. 
+    are prefix of other keys. 
 
-    Like would ``list_dir('mydir')`` returns: 
-       - ``path1``
-       - ``path2``
-       - ``path3/``
-       - ``path4/``
+    Like would ``list_dir('/meta/mydir')`` returns: 
+       - ``/meta/path1``
+       - ``/meta/path2``
+       - ``/meta/path3/``
+       - ``/meta/path4/``
 
-    Thus we know that path1, and path2 are terminal objects with data, and that
-    ``path3`` and ``path4``
+    Thus we know that ``path1``, and ``path2`` are terminal objects with data,
+    and that ``/meta/path3`` and ``/meta/path4``.
 
 
-    Similar to ``ls -p``
+    This is similar to ``ls -p`` on Unix systems.
 
-    Note: In practice I believe this means that this means keys always ends in
-    ``/``, ``.json``, ``.array``, ``.group``, or otherwise they will be chunks
+    Note: In practice this means that this means most returned keys always ends in
+    ``/``, ``.json``, ``.array``, ``.group``, they will otherwise be chunks
     data.
 
 
@@ -1305,6 +1316,9 @@ and ".array". The data key for array chunks is formed by concatenating
 "data", `P`, "/", and the chunk identifier as defined by the chunk
 grid layout.
 
+To get the path ``P`` from a key, either remove the trailing ``.array`` or
+``.group`` as well as the ``meta/root`` prefix.
+
 For example, for an array at hierarchy path "/foo/baz", the
 corresponding metadata key is ``meta/root/foo/baz.array``. If the array
 has two dimensions and a regular chunk grid, the data key for the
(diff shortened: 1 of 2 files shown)


Carreau:
Will play with the open/close buttons, I've pushed many things but they sem to not appear.

Carreau:
Explain in the spec why it's safe why appending `.array`/ `.group` is safe, and what append if we try to store an array or a group which **name** is `foo.group`, or `.array`, as well as how to go from keys to name. 

Also prevent names from being empty.

docs/protocol/core/v3.0.rst
```diff
@@ -269,6 +293,8 @@ node names:
 * must not be a string composed only of period characters, e.g. "." or
   ".."
 
+TODO: likely want to reserver names that start with just a single ``.``
```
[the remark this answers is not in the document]
alimanfoo: FWIW I don't think we need to reserve any names.

docs/protocol/core/v3.0.rst
```diff
@@ -1124,11 +1397,31 @@ Let "+" be the string concatenation operator.
...
+
+
+
 
 **Determine if a node exists**
 
-    @@TODO
+    To determine if a node exists at path `P`, perform `exist()` 
+    exists is tricky, you can query `meta/P.*`, as P might be implicit, 
+    so you need to query both `meta/P.*` and `meta/P/*`.
```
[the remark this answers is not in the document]
alimanfoo: In [zarrita](https://github.com/alimanfoo/zarrita/blob/master/zarrita.py#L323) I try a sequence of operations to determine if a node exists at hierarchy path `P`:

* ``get("meta/root{P}.array")`` - if returns a value, found array 
* ``get("meta/root{P}.group")`` - if returns a value, found explicit group
* ``list_dir("meta/root{P}/")`` - if results, found an implicit group

If all fail, no node exists at path `P`.

docs/protocol/core/v3.0.rst
```diff
@@ -307,89 +333,111 @@ identify a data type.
...
      - 4
      - big-endian
-   * - `>f8`
+   * - ``>f8``
      - double precision float: sign bit, 11 bits exponent, 52 bits mantissa
      - 8
      - big-endian
+   * - ``r*`` (Optional)
+     - raw bits/bytes,  use for extension type fallbacks
+     - variable, given by ``*``, multiple of 8
```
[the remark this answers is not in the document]
Carreau: We'll discuss that today in the spec, my main questions was whether we want that to be the number of bits, or the number of bytes.

If it's number of Bytes, it can likely be any int.
The question is if we allow bits for types shorter the one byte (categorical or boolean), and define a behavior to pad/truncate.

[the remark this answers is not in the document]
jakirkham: @QuLogic, it would be nice to get your thoughts on this (and other bit related points). Would like to make sure this makes sense for architectures that you are thinking about 🙂

docs/protocol/core/v3.0.rst
```diff
@@ -847,35 +981,62 @@ chunking as above, but using an extension data type::
...
-The entry point metadata document must be encoded as JSON. The array and
-group metadata documents must be encoded as per the MIME type given in
+The entry point metadata document must be encoded as JSON. The array (``*.array``s) and
+group metadata documents (``*.group``s) must be encoded as per the MIME type given in
 the ``metadata_encoding`` field in the entry point metadata document
 (described below).
 
 @@TODO more explanation needed?
+Todo questions : How frequent are non-JSON metadata documents ? and how store specific are those. 
+If we allow arbitrary mimetypes, does that precude the sections on arrays and group metadata to be normative ?
```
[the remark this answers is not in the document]
Carreau: Maybe use URI. 
Woudl also allow to encode institution.

docs/protocol/core/v3.0.rst
```diff
@@ -626,9 +716,9 @@ containing the following names:
 
 ``metadata_encoding``
 
-    A string containing the MIME type that has been used for encoding
+    A string containing the URI that has been used for encoding
     group and array metadata documents. If JSON encoding is used, this
-    value must be "application/json".
+    value must be ``"https://purl.org/zarr/spec/metadata/json/3.0"``.
```
alimanfoo, line 721: In general we want URIs to dereference to some documentation that defines the metadata encoding. But for JSON, I don't think we need to create a new URI, we could just point to an existing URI. E.g., we could use the JSON spec URI: `https://tools.ietf.org/html/rfc8259`.

Carreau: That is fine with me, but one of the reason we discuss to use URI instead of mimetype was the the same mimetype can encore metadata in a different way. So pointing to the JSON spec URI seem to me like it has the same issue.

You do would to not only point to the format of the file, but also it's structure; At least that was my understanding.

docs/protocol/core/v3.0.rst
```diff
@@ -307,89 +331,128 @@ identify a data type.
...
 
+Additionally to these base types, an implementation should also handle the
+raw/opaque pass through type designated by the lowercase letter ``r`` followed
+by the number of bits. For example, ``r3``, ``r11`` should be understood as
+fallback types of respectively 3 and 11 bytes longs, and implementation should
+ensure to pass or accept those types unchanged when using extension fallback.
+For array whose total length is not a multiple of 8, zarr implementation can
+expect expects a buffer which length is a multiple of 8bits, padded with
+appropriate value at the end, and will in the same way return buffers padded to
+the next multiple of 8 bytes.
```
QuLogic, line 432: This paragraph confuses bits and bytes a lot. First it says bits, then it says bytes, then bits again, then bytes. I think they should all say bits.

docs/protocol/core/v3.0.rst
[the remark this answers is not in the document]
QuLogic: I think generally all systems in Fedora use 8-bit bytes, so there is no worry there. The key issue is order. I see there are definitions for endian order on multi-byte values, but I don't see anything about bit order (for a `packbits`-equivalent type, it looks like). If there's padding, is it between every entry, or just at the end? If the number of bits is > 8, is it first byte, then second, or the other way? Are the >8 bits in the lower or upper bits of the next byte? This is somewhat explained in examples, but not exactly explicit.

docs/protocol/core/v3.0.rst
Carreau: bits everywhere; sorry, thanks.

docs/protocol/core/v3.0.rst
```diff
@@ -52,6 +54,25 @@ or by making a pull request against the
...
+storage have characteristics, capabilities and usage patterns that can widely
+differ from the assumptions of spec v2. V3 is design to consider online
+stores in particular we want to achieve the following.
+
+ - No assumption that the underlying store have locking ability.
+ - Ability to do concurrent writes with the assumption that writes from clients will be consistent, but non atomic.
+
+
+Unlike Zarr spec version 2, the spec V3 have mainly the following difference:
+  - V3 is a flat key-value store instead of an hierarchical store. Hierarchy is implied.
```
joshmoore, line 71: What are the implications of this on the local filesystem implementations? cc: @jni

docs/protocol/core/v3.0.rst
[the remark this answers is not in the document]
Carreau: Padding would just be at the end; not in between. I have no objections in changing. I think whether the order is lower first or lower last would be up to the extensions, or library using this which actually know the internal types. `r` is mostly meant as a pass through for when you can't fallback on an another of the type you can fallback on. Though I don't have that much experience with different endianness, so if you think this should be explicit I'll trust you.

docs/protocol/core/v3.0.rst
Carreau: Not much:
 - you can't open a hierarchy at an arbitrary point, you have to open the root. (zarr_open, could walk up the tree until is finds the root, and return a subgroup, that's not an issue, but that's just user facing).
 - meta/ and data/ are in two subtrees, so the storage format is one layer deeper than v2

Also the chink separator is now `/` by default; which has a much bigger impact on the disk layout as now you have a deeper hierarchy, with less files per folder.

jni: > Also the chink separator is now / by default; which has a much bigger impact on the disk layout as now you have a deeper hierarchy, with less files per folder.

Ah, that's the crux. This particular sentence makes it sound like the opposite is true, but they are talking about different things. Thanks for clarifying @Carreau! I'm not sure whether it makes sense to add a similar clarification with this bullet point — I'll leave that to people more familiar with this whole ecosystem. =)

docs/protocol/core/v3.0.rst
```diff
@@ -665,7 +802,7 @@ ignored if not understood::
 
     {
         "zarr_format": "https://purl.org/zarr/spec/protocol/core/3.0",
-        "metadata_encoding": "application/json",
+        "metadata_encoding": "https://tools.ietf.org/html/rfc8259",
         "extensions": [
```
Carreau, line 806: let's go with `https://purl.org/zarr/spec/protocol/core/3.0` + anchor to metadata section.

docs/protocol/core/v3.0.rst
```diff
@@ -307,89 +341,126 @@ identify a data type.
...
+ensure to pass or accept those types unchanged when using extension fallback.
+For array whose total length is not a multiple of 8 bits, Zarr implementation can
+expect a buffer which length is a multiple of 8 bits, padded with
+appropriate value at the end, and will in the same way return buffers padded to
+the next multiple of 8 bits. 
+
+For example, here is a array of 5 elements of opaque type ``r3`` will be stored as
+
+ -  ``AAABBBCC CDDDEEE_``, 3*5 = 15 bits, rounded to 2 bytes, with arbitrary
+   last bit.
```
Carreau, line 447: "bullet list ends w/o a blank line".

docs/protocol/core/v3.0.rst
```diff
@@ -314,82 +359,111 @@ Core data types
...
      - 4
      - big-endian
-   * - `>i8`
+   * - ``>i8``
      - signed integer
      - 8
      - big-endian
-   * - `u1`
+   * - ``u1``
      - signed integer
```
davidbrochart, line 395: unsigned integer?

docs/protocol/core/v3.0.rst
```diff
@@ -314,82 +359,111 @@ Core data types
...
      - big-endian
-   * - `>f8`
+   * - ``>f8``
      - double precision float: sign bit, 11 bits exponent, 52 bits mantissa
      - 8
      - big-endian
+   * - ``r*`` (Optional)
+     - raw bits,  use for extension type fallbacks
+     - variable, given by ``*``, is limited to be a multiple of 8.
+     - N/A
```
davidbrochart, line 437: I think this is too confusing to describe it as bits. The size column is in bytes, and we are saying it is limited to be a multiple of 8 bits anyway. I think we should just say "raw bytes".

davidbrochart: Another reason is that all the other identifiers have a value in number of bytes, e.g. `i1` which is a 1-byte integer.

docs/protocol/core/v3.0.rst
```diff
@@ -52,6 +52,43 @@ or by making a pull request against the
...
 <https://github.com/orgs/zarr-developers/teams/core-devs>`_.
 
+Goal of v3 spec and main difference from v2
+===========================================
+
+Zarr spec v2 was originally designed around local filesystem, but Zarr has
+grown and is now regularly deployed on cloud / object storage. Those kind of
+storage have characteristics, capabilities and usage patterns that can widely
+differ from the assumptions of spec v2. V3 is design to consider online
+stores in particular we want to achieve the following.
```
davidbrochart, line 62:
```suggestion
differ from the assumptions of spec v2. V3 is designed to consider online
stores, in particular we want to achieve the following:
```

docs/protocol/core/v3.0.rst
```diff
@@ -52,6 +52,43 @@ or by making a pull request against the
 This document was produced by the `Zarr core development team
 <https://github.com/orgs/zarr-developers/teams/core-devs>`_.
 
+Goal of v3 spec and main difference from v2
```
davidbrochart, line 55:
```suggestion
Goal of v3 spec and main difference with v2
```

docs/protocol/core/v3.0.rst
```diff
@@ -52,6 +52,43 @@ or by making a pull request against the
...
+===========================================
+
+Zarr spec v2 was originally designed around local filesystem, but Zarr has
+grown and is now regularly deployed on cloud / object storage. Those kind of
+storage have characteristics, capabilities and usage patterns that can widely
+differ from the assumptions of spec v2. V3 is design to consider online
+stores in particular we want to achieve the following.
+
+ - No assumption that the underlying store have locking ability.
+ - Ability to do concurrent writes with the assumption that writes from clients will be consistent, but non atomic.
```
davidbrochart, line 65:
```suggestion
 - No assumption that the underlying store has locking ability.
 - Ability to do concurrent writes with the assumption that writes from clients will be consistent, but not atomic.
```

docs/protocol/core/v3.0.rst
```diff
@@ -52,6 +52,43 @@ or by making a pull request against the
...
+
+
+Unlike Zarr spec version 2, the spec V3 have mainly the following difference:
+  - V3 is a flat key-value store instead of an hierarchical store. Hierarchy is implied.
+  - V3 has an explicit root, while v2 roots and group could not be distinguished.
```
davidbrochart, line 70:
```suggestion
Unlike Zarr spec v2, the spec v3 has mainly the following differences:
  - V3 is a flat key-value store instead of a hierarchical store. Hierarchy is implied.
  - V3 has an explicit root, while v2 roots and groups could not be distinguished.
```

docs/protocol/core/v3.0.rst
```diff
@@ -52,6 +52,43 @@ or by making a pull request against the
...
+  - Separation of the data and  metadata key space.
+  - Explicit support for extensions.
+  - chunk separator is ``/`` by default.
+
+This means that a store cannot be opened at an arbitrary point, and need to be
+opened at the root. User facing convenience functions could walk up a given
```
davidbrochart, line 76:
```suggestion
This means that a store cannot be opened at an arbitrary point, but needs to be
opened at the root. User facing convenience functions could walk a given
```

docs/protocol/core/v3.0.rst
Carreau: Sure, but then we'll never be able to go lower or non multiple of 8.

docs/protocol/core/v3.0.rst
Carreau: probably my fault when I typed a vim macro. 

I gave you access to my fork if you want to make changes directly.

Carreau:
I've revised the spec with respect to todays conversation.

docs/protocol/core/v3.0.rst
```diff
@@ -542,13 +542,13 @@ is contained within the chunk at grid index (1, 7, 2) and has coordinates
 
 The identifier for chunk with grid index (``i``, ``j``, ``k``, ...) is
 formed by joining together ASCII string representations of each index
-using a separator. The default value for the separator is the period
+using a separator and prefixed with the string `'c'`. The default value for the separator is the period
 character ( by default ``/``), but this may be configured by providing a ``separator``
```
davidbrochart:
```suggestion
using a separator and prefixed with the string `'c'`. The default value for the separator is the slash
character (by default ``/``), but this may be configured by providing a ``separator``
```

Carreau:
From today's meeting we are going to merge that in order to make further changes smaller and easier to digest. Thanks everyone.

Issue #57, opened 2020-03-31
Node name case sensitivity


alimanfoo:
Currently the [v3.0 core protocol draft section on node names](https://zarr-specs.readthedocs.io/en/core-protocol-v3.0-dev/protocol/core/v3.0.html#node-names) states the following:

> Note that node names are used to form storage keys, and that some storage systems will perform a case-insensitive comparison of storage keys during retrieval. Therefore, within a hierarchy, all nodes within a set of sibling nodes must have a name that is unique under case-insensitive comparison. E.g., the names “foo” and “FOO” are not allowed for sibling nodes.

This constraint, however, is problematic in scenarios where nodes are being created in parallel. This is because requiring that sibling nodes have a unique name under case-insensitive comparison would require that a check is made before node creation to ensure no name collisions, and any such check would require synchronisation of node creation at least within the same group.

Currently a design goal for the spec is to avoid constraints which require operations to be synchronised, thus allowing multiple parallel processes to work simultaneously, including creating nodes within the same hierarchy. This avoids implementation complexity (no need for locks or other synchronisation mechanisms) and also works better on stores with eventual consistency behaviour.

I suggest we remove this constraint from the core protocol. I.e., implementations of the core protocol **are not** expected to check for case-insensitive node name collisions. 

However, instead we add a usage note, which provides some guidance for implementing applications using a zarr protocol implementation, including recommending that applications avoid creating sibling nodes with case-insensitive name collisions wherever that is possible. 

I.e., we push this to the application layer.

Carreau:
> However, instead we add a usage note, which provides some guidance for implementing applications using a zarr protocol implementation, including recommending that applications avoid creating sibling nodes with case-insensitive name collisions wherever that is possible.

Then do you want to add on the spec that stores must be able to handle case sensitive names ? In the spec it says `Note that node names are used to form storage keys` it does not say _how_, and that might push for example filesystems to not just use a key but another scheme.

alimanfoo:
Hi @Carreau,

The [subsection on storage keys](https://zarr-specs.readthedocs.io/en/core-protocol-v3.0-dev/protocol/core/v3.0.html#storage-keys) describes how keys are constructed from node paths. 

It is then up to a store spec to decide how to use these keys and store data. 

[...]

PR #85, opened 2020-07-15
Node name changes

alimanfoo:
Following on from discussion on the community call today, this PR proposed two changes regarding the specification of node names:

* The spec now requires that stores are case sensitive. E.g., the operations `set('foo', a)` and `set('FOO', b)` will result in two distinct (key, value) pairs being stored. Subsequently `get('foo')` will return *a* and `get('FOO')` will return *b*. 

* A note has been added stating our desire to support Unicode in node names and linking to the appropriate issue for comment.

xref #56, #57

### DIFF
--- a/docs/protocol/core/v3.0.rst
+++ b/docs/protocol/core/v3.0.rst
@@ -962,7 +962,7 @@ which do not contain the character "/" after the given prefix.
 
 Note that because keys are case sensitive, it is assumed that the operations 
 ``set("foo", a)`` and ``set("FOO", b)`` will result in two separate (key, value) 
-pairs being stored. Subsequently ``get("foo")`` will return *a* and ``get("bar")`` 
+pairs being stored. Subsequently ``get("foo")`` will return *a* and ``get("FOO")`` 
 will return *b*. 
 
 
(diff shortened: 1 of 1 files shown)


alimanfoo:
cc @Carreau

jakirkham:
I think this came up on a previous call. On macOS, the filesystem is case insensitive by default (and this is the common case in the wild). Unfortunately changing a filesystem to be case sensitive is non-trivial as it requires reformatting. Also it may cause breakage in applications that have already assumed things are not case sensitive.

In practice, if neither `FOO` or `foo` existed and one of them was created, casing would match the one created (say `FOO`). However there would not be an option to create both as they would resolve to the same file or directory.

jakirkham:
Should add I'm not saying we should make the choice based on macOS's behavior, which is admittedly odd. Just noting this will be a stumbling block for a potentially large set of users. Perhaps we can recommend users avoid creating a path that would conflict on case insensitive storage mediums? Though leave this in users hands.

Carreau:
Closed in favor of #97 which is rebased to solve conflicts, but has the same content.

PR #98, opened 2020-10-12
Metadata Key suffixes

### LINKED ISSUE #63: Support suffix for metadata files
In the current v3 core protocol draft, keys for metadata documents have no suffix to represent the document format, they are just keys like "meta/root/foo.group" and "meta/root/foo/bar.array". In some situations (e.g., where file systems or web servers are being used as storage) it would be useful to have a suffix that indicates the format, i.e., ".json" for the default. Consider adding support for this.

Carreau:
Closes #63

The rational is that backends may be explored by non zarr-aware tools,
like a filesytem and a GUI browser like Finder/Explorer/Nautilus, and
the `.json` extension will help to view/edit json in those.

While it may seem like this shoudl be store specific, once again the
fact that a zarr-hierarchy can be copied from one storage to another by
non-zarr-aware tool indicate that this shoudl be handled above the sotre
level.

### DIFF
--- a/docs/protocol/core/v3.0.rst
+++ b/docs/protocol/core/v3.0.rst
@@ -736,6 +736,16 @@ containing the following names:
     For document using the default JSON encoding and format describe in this document
     then the value must be ``"https://purl.org/zarr/spec/protocol/core/3.0``.
 
+``metadata_key_suffix``
+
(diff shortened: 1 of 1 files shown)


alimanfoo (approved):

alimanfoo:
Thanks @Carreau.

PR #100, opened 2020-10-21
Replace tab by spaces and clarify TODO point

Carreau:


alimanfoo:
Thanks @Carreau. Just to surface the comment from the community call today, I think it would go well to pull this kind of information out into a separate "Implementation notes" section. That would give a bit more room to explain these points in more detail.

PR #95, opened 2020-09-29
Add initial draft for blosc codec

davidbrochart:
I have many questions:
- should we provide a compression algorithm name, as in [numcodecs](https://numcodecs.readthedocs.io/en/stable/blosc.html#module-numcodecs.blosc)?
- there doesn't seem to be a reference document for Blosc. Is the reference the C implementation GitHub URL?
- it looks like Blosc and Blosc2 are incompatible. Should we provide the version number in the meta data?
- is the file format the raw storage of the memory compressed data?

PR #102, opened 2020-10-21
Refactor codec specs into a single doc

alimanfoo:
On reflection, I thought it might be better to refactor the codec specifications into a single document, because this would reduce the overhead of adding new codecs and maintaining the documentation. Generally for each codec we are citing some existing documentation elsewhere for the definition of the encoding algorithm and format, and so we don't actually need to provide much information for each codec, hence having a single document per codec seems overkill.

So this PR brings the existing gzip codec spec and the proposed blosc codec spec (#95) into a single "Codec registry" specification.

Comments welcome.


alimanfoo:
cc @davidbrochart

docs/codecs.rst
```diff
@@ -1,11 +1,207 @@
...
+.. [RFC1951] P. Deutsch. DEFLATE Compressed Data Format Specification version
+   1.3. Requirement Levels. May 1996. Informational. URL:
+   https://tools.ietf.org/html/rfc1951
+
+.. [RFC1952] P. Deutsch. GZIP file format specification version 4.3.
+   Requirement Levels. May 1996. Informational. URL:
+   https://tools.ietf.org/html/rfc1952
+
+.. [BLOSC] F. Alted. Blosc Chunk Format. URL:
+   https://github.com/Blosc/c-blosc/blob/master/README_CHUNK_FORMAT.rst
```
davidbrochart, line 197: This document describes the "Blosc Chunk Format". Does it mean a Zarr chunk consists of one or more Blosc chunks? If so, is the Zarr chunk a simple concatenation of the Blosc chunks?

docs/codecs.rst
```diff
@@ -1,11 +1,207 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+cname:
+    A string identifying the internal compression algorithm to be
+    used. At the time of writing, the following values are supported
+    by the c-blosc library: "lz4", "lz4hc", "blosclz", "zstd",
+    "snappy", "zlib".
+    
+clevel:
+    An integer between 1 and 9 indicating the compression level.
```
davidbrochart, line 131: According to the [c-blosc source file](https://github.com/Blosc/c-blosc/blob/master/blosc/blosc.h#L147-L148), the compression level can also be 0, in which case there is no compression.
I think we should copy/paste the description of the `level` parameter from GZip, as the relation between compression level and speed of compression probably applies to every underlying compression algorithm.
```suggestion
    An integer from 0 to 9 which controls the speed and level of
    compression. A level of 1 is the fastest compression method and
    produces the least compressions, while 9 is slowest and produces
    the most compression. Compression is turned off completely when
    level is 0.
```

docs/codecs.rst
```diff
@@ -1,11 +1,207 @@
...
+
+shuffle:
+    An integer value in the set {0, 1, 2, -1}. A value of 1
+    indicates that byte-wise shuffling is performed in addition to
+    compression. A value of 2 indicates the bit-wise shuffling is
+    performed in addition to compression. If a value of -1 is given,
+    then default shuffling is used: bit-wise shuffling for buffers
+    with item size of 1 byte, byte-wise shuffling otherwise.
+    Shuffling is turned off completely when the value is 0.
```
davidbrochart, line 140: Instead of "in addition to compression", should we say "prior to compression"? It seems more accurate.
We might also mention that shuffling "may improve performance and/or compression ratio" (as stated in https://blosc.org/pages).
```suggestion
    An integer value in the set {0, 1, 2, -1} indicating the way
    bytes or bits are rearranged, which can lead to faster
    and/or greater compression. A value of 1
    indicates that byte-wise shuffling is performed prior to
    compression. A value of 2 indicates the bit-wise shuffling is
    performed prior to compression. If a value of -1 is given,
    then default shuffling is used: bit-wise shuffling for buffers
    with item size of 1 byte, byte-wise shuffling otherwise.
    Shuffling is turned off completely when the value is 0.
```

docs/codecs.rst
```diff
@@ -1,11 +1,207 @@
...
+
+blocksize:
+    An integer giving the size in bytes of blocks into which a
+    buffer is divided before compression.
```
davidbrochart, line 144: Should we mention that when `0` is passed, [an automatic size is used](https://github.com/Blosc/c-blosc/blob/master/blosc/blosc.h#L238-L239)?
```suggestion
    An integer giving the size in bytes of blocks into which a
    buffer is divided before compression. A value of 0
    indicates that an automatic size will be used.
```

docs/codecs.rst
```diff
@@ -1,11 +1,207 @@
...
+    produces the least compressions, while 9 is slowest and produces
+    the most compression. Compression is turned off completely when
+    level is 0.
+
+For example, the array metadata below specifies that the compressor is
+the Gzip codec configured with a compression level of 1::
+
+    {
+        "compressor": {
+            "codec": "https://purl.org/zarr/spec/codecs/gzip",
```
davidbrochart, line 97: The spec version isn't part of the URL anymore? We previously had:
```suggestion
            "codec": "https://purl.org/zarr/spec/codecs/gzip/1.0",
```

docs/codecs.rst
```diff
@@ -1,11 +1,207 @@
...
+    buffer is divided before compression.
+
+For example, the array metadata document below specifies that the
+compressor is the Blosc codec configured with a compression level of
+1, byte-wise shuffling, the ``lz4`` compression algorithm and the
+default block size::
+
+    {
+        "compressor": {
+            "codec": "https://purl.org/zarr/spec/codecs/blosc",
```
davidbrochart, line 153: The spec version isn't part of the URL anymore? We previously had:
```suggestion
            "codec": "https://purl.org/zarr/spec/codecs/blosc/1.0",
```

davidbrochart:
Thanks @alimanfoo, it is much better than what we had previously with on document per codec, and I like the way parameters are described, similar to what we are used to in a Python docstring.

docs/codecs.rst
alimanfoo: The "Blosc Chunk Format" describes how blosc encodes an input buffer. So one zarr chunk becomes one blosc chunk once encoded.

docs/codecs.rst
alimanfoo: Yes, see above.

Note that Blosc 2 would be treated as an entirely new codec, and we would probably give it a URI like `https://purl.org/zarr/spec/codecs/blosc2`.

alimanfoo:
Thanks @davidbrochart, I've added your suggestions.

Regarding the codec URIs, are you happy that we remove the version numbers, or would you like to discuss this some more?

docs/codecs.rst
davidbrochart: :+1:

docs/codecs.rst
davidbrochart: :+1:

docs/codecs.rst
davidbrochart: :+1:

davidbrochart:
> Regarding the codec URIs, are you happy that we remove the version numbers, or would you like to discuss this some more?

I'm happy like that, I was indeed confused if it was the version of the spec or the version of e.g. the GZip library.

Carreau (approved):

joshmoore:
I've attempted an update of this PR and intend to merge it into [core-protocol-v3.0-dev](https://github.com/zarr-developers/zarr-specs/tree/core-protocol-v3.0-dev) in case anyone would like to re-review.

PR #104, opened 2020-10-23
Review

davidbrochart:


davidbrochart:
What does the following mean @Carreau @alimanfoo?
```
An array element may be empty, or it may have a value.
```

davidbrochart:
The difference between `compressor` and `codec` is still not clear to me. I understand that a compressor has a codec and its associated configuration, so basically a compressor is an "instance of a codec"?
cc @alimanfoo @Carreau

alimanfoo:
> What does the following mean @Carreau @alimanfoo?
> 
> ```
> An array element may be empty, or it may have a value.
> ```

This is just meant to convey that an array may exist but may be empty, i.e., may not have had some or all of its values assigned. Does that make sense? Suggestions welcome to clarify.

alimanfoo:
> The difference between `compressor` and `codec` is still not clear to me. I understand that a compressor has a codec and its associated configuration, so basically a compressor is an "instance of a codec"?
> cc @alimanfoo @Carreau

A codec is something that implements the codec interface, which we define as a pair of functions, encode and decode. 

A codec may be used as a compressor. 

(A codec may also be used as a filter. Filters are supported in zarr v2 but moved to a [protocol extension in zarr v3](https://zarr-specs.readthedocs.io/en/core-protocol-v3.0-dev/protocol/extensions/filters/v1.0.html).)

Does that make sense?

davidbrochart:
> This is just meant to convey that an array may exist but may be empty, i.e., may not have had some or all of its values assigned. Does that make sense? Suggestions welcome to clarify.

Then I would rephrase with something like:
```
An array may not have been fully initialized.
```
That also raises the question of uninitialized chunks, which may not even exist on the file system. When an element in the uninitialized chunk is read, a default value is used. I may have missed that, but is it specified, or does it just not matter?

davidbrochart:
> A codec is something that implements the codec interface, which we define as a pair of functions, encode and decode.
> 
> A codec may be used as a compressor.
> 
> (A codec may also be used as a filter. Filters are supported in zarr v2 but moved to a [protocol extension in zarr v3](https://zarr-specs.readthedocs.io/en/core-protocol-v3.0-dev/protocol/extensions/filters/v1.0.html).)
> 
> Does that make sense?

Thanks @alimanfoo, I may have to read more about that.

davidbrochart:
> That also raises the question of uninitialized chunks, which may not even exist on the file system. When an element in the uninitialized chunk is read, a default value is used. I may have missed that, but is it specified, or does it just not matter?

OK, this is the purpose of the `fill_value` field.

docs/protocol/core/v3.0.rst
```diff
@@ -1422,7 +1422,7 @@ Let "+" be the string concatenation operator.
 **Discover all nodes in a hierarchy**
 
     To discover all nodes in a hierarchy, one can call
-    ``list("meta/")``. All keys represent either explicit group or
+    ``list_prefix("meta/root/")``. All keys represent either explicit group or
```
davidbrochart, line 1425: @alimanfoo `list` doesn't accept any parameter, so I guess we should use `list_prefix`, and I guess that `/root` was missing?

davidbrochart:
@alimanfoo for the [abstract store interface](https://github.com/zarr-developers/zarr-specs/blob/core-protocol-v3.0-dev/docs/protocol/core/v3.0.rst#abstract-store-interface), I was wondering if the operations (`get`, `set`, `delete`...) are required to be methods of the store object, or free functions, or if we leave that as an implementation detail.

Carreau:
> @alimanfoo for the [abstract store interface](https://github.com/zarr-developers/zarr-specs/blob/core-protocol-v3.0-dev/docs/protocol/core/v3.0.rst?rgh-link-date=2020-11-02T17%3A23%3A48Z#abstract-store-interface), I was wondering if the operations (`get`, `set`, `delete`...) are required to be methods of the store object, or free functions, or if we leave that as an implementation detail.

I think as you pointed out that it should be an implementation details, for some language you probably don't even have the concept of methods on objects.

docs/protocol/core/v3.0.rst
Carreau: I think maybe just `list()` ? 

```
``list`` - Retrieve all `keys` in the store.

    | Parameters: none
    | Output: set of `keys`
```

Or do we get rid of list and say that it should be constructed from list_prefix ?

davidbrochart:
Sounds good @Carreau, thanks.
Also, the `delete` operation might be problematic in C++, since it has a particular meaning. I understand that it is the natural way of e.g. removing an item from a dictionary in Python, and maybe it was inspired by that, but maybe we could think of different name, like `erase`?

docs/protocol/core/v3.0.rst
```diff
@@ -1383,16 +1383,16 @@ Let "+" be the string concatenation operator.
...
     according to the information in the array metadata stored under
     the key ``array_meta_key(P)``.
 
 **Retrieve element values in an array**
 
     To retrieve element in an array at path `P` and coordinate (`i`,
-    `j`, ...)  perform ``get(data_key(P, i, j, ...), value)``, where
-    `value` is the serialisation of the corresponding chunk, encoded
+    `j`, ...), perform ``get(data_key(P, i, j, ...))``. The returned
+    value is the serialisation of the corresponding chunk, encoded
```
davidbrochart, line 1395: It was probably due to a copy/past from above, since `get` *returns* the value.

davidbrochart:
@alimanfoo @Carreau OK to merge?

Carreau (approved):

Carreau:
FIne with me.

davidbrochart:
> maybe we could think of different name, like `erase`?

What do you think about changing `delete` to `erase`? We might also want to change `delete_prefix` to `erase_prefix` for consistency.

docs/protocol/core/v3.0.rst
davidbrochart: I actually implemented `list()` as `list_prefix("")`, but it's also good to keep `list()` as a shortcut.
But we can't use `list()` to discover all the nodes in a hierarchy, since we need to prefix with `"meta/root/"`.

rabernat:
Hi @davidbrochart -- your suggestions look great. I unfortunately created a merge conflict when I merged #105, but it doesn't look very difficult to resolve.

davidbrochart:
No worries @rabernat, I just rebased.

docs/protocol/core/v3.0.rst
```diff
@@ -144,26 +144,28 @@ Questions that still need to be resolved
...
- - extensions and ``must_understand = True`` might be too restrictive. Work a draft implementation with extensions and
-   see how far we can go. List of extensions to implement: 
-   
-      - Boolean
-      - Complex
-      - Datetime
-      - Named dimensions
-      - Awkward arrays
-        
+ - https://github.com/zarr-developers/zarr-specs/issues/72 to potentially split
```
rabernat, line 147:
```suggestion
 - Should core metadata and user attributes be stored together or separate documents? ([GH72](https://github.com/zarr-developers/zarr-specs/issues/72)) 
```

docs/protocol/core/v3.0.rst
```diff
@@ -144,26 +144,28 @@ Questions that still need to be resolved
...
+   large metadata documents.
+ - extensions and ``must_understand = True`` might be too restrictive. Work a
+   draft implementation with extensions and
```
rabernat, line 150: I don't understand the syntax "work a draft implementation"? Do you mean "develop a draft implementation"?

docs/protocol/core/v3.0.rst
```diff
@@ -144,26 +144,28 @@ Questions that still need to be resolved
...
+   see how far we can go. List of extensions to implement:
```
rabernat, line 151:
```suggestion
   see how far we can go. Possible list of extensions to implement:
```

docs/protocol/core/v3.0.rst
```diff
@@ -769,12 +783,12 @@ name/value pairs. This section also defines how metadata documents are
...
-can be used to defined other format to array-level and group-level metadata
-document; in the case where non-json metadata document are use in a zarr
-hierarchy the following sections on group and array level metadata are
-non-normative; but other metadata format as expected to define some equivalence
-relations with the JSON documents.
+Only the top level metadata document ``zarr.json`` is guaranteed to be of JSON
+type, and can be used to define other formats for array-level and group-level
+metadata documents. In the case where non-JSON metadata documents are used in a
+Zarr hierarchy, the following sections on group and array level metadata are
+non-normative; but other metadata formats are expected to define some
```
rabernat, line 790:
```suggestion
non-normative, but other metadata formats are expected to define some
```

docs/protocol/core/v3.0.rst
```diff
@@ -1484,16 +1501,15 @@ Let "+" be the string concatenation operator.
...
-   array metadata ``data_type`` keys.
- - Arrays Extensions – Non rectilinear grids, and
-   variable length types.
+ - Core data type extensions – for example adding the ability to store fixed size
+   types such as complex or datetime in chunks. These are directly declared in the
+   array metadata ``data_type`` key.
+ - Arrays extensions – non rectilinear grids, and variable length types.
 
-There are no group extensions as as Zarr v3.0
+There are no group extensions as in Zarr v3.0.
```
rabernat, line 1512:
```suggestion
There are no group extensions in Zarr v3.0.
```

rabernat (commented):
Just some minor typos.

docs/protocol/core/v3.0.rst
davidbrochart: My change is was just on the form. @Carreau might have written it?

davidbrochart:
> Just some minor typos.

Thanks Ryan. I'm wondering if it's time to get rid of these notes here and there, that are not intended to be part of the document?

rabernat:
> I'm wondering if it's time to get rid of these notes here and there, that are not intended to be part of the document?

I'm not sure what notes you're referring to.

I'm going to merge this so I can move forward with my review as a new PR.

Issue #101, opened 2020-10-21
Preparation for core protocol v3 first public working draft and RFC


alimanfoo:
This issue is intended for planning and coordination towards publishing a First Public Working Draft of the Zarr core protocol v3 specification and issuing a public request for comments (RFC).

We discussed this on today's [community call](https://docs.google.com/document/d/1o-xSVe1iAKNIfz2a9U0-90-z96XwwTT9jzp-ZuyOJQc/edit?usp=sharing), we suggested to aim for publishing the spec and issuing the RFC on **Wednesday 4 November**.

## Suggested plan

* Ask Zarr core devs to read through the [current editor's draft](https://zarr-specs.readthedocs.io/en/core-protocol-v3.0-dev/protocol/core/v3.0.html) and raise any issues they think need to be addressed before we can issue an RFC. N.B., the spec doesn't have to be perfect before RFC, so the goal here is just to deal with any obvious mistakes, inconsistencies, potential confusions, major gaps etc.
* Spec editors (@Carreau, @alimanfoo) resolve any issues raised above.
* Spec editors make a final PR to change the document type from "Editor's Draft" to "First Public Working Draft" and merge into the master branch on the zarr-specs repo.
* Update the RTFD config for zarr-specs.readthedocs.io to serve from the master branch (currently serving from dev branch).
* Write a blog post for the [zarr blog](https://zarr-developers.github.io/) declaring the spec is published, announcing the RFC, and providing any other useful information like motivation for this spec, pointers to current implementation work, plan for future timeline, etc.
* Publicize the RFC via Twitter (@zarr_dev, ...) and other channels

## Notes

* The plan is to ask for comments to be submitted by raising issues on the zarr-specs repo.
* It's suggested we follow a convention similar to W3C where they label a spec as "Editor's Draft" if it is an internal draft only intended for consumption within the group of people actively working on it, then will publish a "Working Draft" when the document has reached some maturity and they want to make it public and request comments from outside the core team. Here "Working Draft" indicates the documents [maturity level](https://www.w3.org/2004/02/Process-20040205/tr.html#maturity-levels). They will also add a date, e.g., "First Public Working Draft 21 October 2020", which then allows for publication of a second working draft if necessary.
* Following on from this, it's suggested that we reserve the master branch on the zarr-specs repo for specs that are intended to be public, i.e., have a "Working Draft" maturity level or beyond. 
* At some point we should capture and document these conventions around process and document maturity levels.

[...]

PR #105, opened 2020-10-23
Intro edit

alimanfoo:
This PR has some edits to the introduction, to try to make the document as accessible as possible to a new reader.

In particular, I've tried to move some more technical matter regarding comparisons with v2 to the end, so the reader can start with a general introduction and motivation, then move into the concepts and protocol.

Comments welcome of course.

### DIFF
--- a/docs/protocol/core/v3.0.rst
+++ b/docs/protocol/core/v3.0.rst
@@ -14,13 +14,12 @@ Issue tracking:
     `GitHub issues <https://github.com/zarr-developers/zarr-specs/labels/core-protocol-v3.0>`_
 
 Suggest an edit for this spec:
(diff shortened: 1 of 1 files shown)


alimanfoo:
cc @davidbrochart

davidbrochart:
LGTM

Carreau (approved):

rabernat (approved):
These look like great suggestions.

Today I sat down to work on my review of v3, but then I noticed this. So I would prefer to merge this first.

rabernat:
Apologies if I overstepped by merging this. Despite having read https://github.com/zarr-developers/zarr-specs/issues/101, I am not 100% clear on the process.

jakirkham:
Seems fine to merge this. We can always do more PRs to refine things further 😄 It seems more important at this stage to get more people comfortable with merging here so we can improve our bandwidth a bit (especially as this work happens asynchronously) 😉

rabernat:
That's what I'm trying to do! :)

PR #106, opened 2020-11-05
Rename delete to erase

davidbrochart:
The `delete` keyword can be problematic in C++. I understand that `delete` was probably inspired by the way we remove an item in a dictionary in Python, but maybe `erase` is a more neutral word across programming languages? `remove` could also be an option.


Carreau:
Renaming sounds good to me, I have no objections.

davidbrochart:
OK to merge?

Issue #116, opened 2021-05-19
v3: https://purl.org/zarr/spec/codec/XXX/1.0: codec singular or codecs plural ?


rouault:
https://github.com/zarr-developers/zarr-specs/blob/core-protocol-v3.0-dev/docs/protocol/core/v3.0.rst mentions in examples ``"codec": "https://purl.org/zarr/spec/codec/gzip/1.0"`` (codec singular) whereas https://github.com/zarr-developers/zarr-specs/blob/core-protocol-v3.0-dev/docs/codecs/gzip/v1.0.rst mentions ``"codec": "https://purl.org/zarr/spec/codecs/gzip/1.0"`` (codec plural).

joshmoore:
Thanks, @rouault. I pushed an update. All the best, ~Josh

rouault:
There was another occurence. Proposed fix in https://github.com/zarr-developers/zarr-specs/pull/117

@alimanfoo  Zarrita use codec singular. It should likely be updated

joshmoore:
> @alimanfoo Zarrita use codec singular. It should likely be updated

That poses an interesting problem:

```
zarr_implementations $ git grep spec/codec
examples/zarrita.zr3/meta/root/blosc/lz4.array.json:        "codec": "https://purl.org/zarr/spec/codec/blosc/1.0",
examples/zarrita.zr3/meta/root/gzip.array.json:        "codec": "https://purl.org/zarr/spec/codec/gzip/1.0",
examples/zarrita.zr3/meta/root/zlib.array.json:        "codec": "https://purl.org/zarr/spec/codec/zlib/1.0",
(base) /opt/zarr_implementations $
```
as well as https://github.com/xtensor-stack/xtensor-zarr/blob/484f98bd06c53e7dc3d1653612907c479954259f/include/xtensor-zarr/xzarr_array.hpp#L40

I took the plural because it was more prevalent in the spec doc, but looks like the singular has made it out into the wild.

davidbrochart:
We can make the change in xtensor-zarr.

joshmoore:
Please see https://github.com/zarr-developers/zarr-specs/pull/119 for the move to the singular version.

cc: @alimanfoo @Carreau

PR #119, opened 2021-06-14
Use singular for "spec/codec"

joshmoore:
Since https://purl.org/zarr/spec/codec/gzip/1.0 is already
in the wild in xtensor-zarr and zarr_implementations, roll-
back the previous choice which was based on a majority of
uses in this repo.

fix: #116

### DIFF
--- a/docs/codecs/gzip/v1.0.rst
+++ b/docs/codecs/gzip/v1.0.rst
@@ -6,7 +6,7 @@ Gzip Codec (version 1.0)
 -----------------------------
 
 Specification URI:
-    https://purl.org/zarr/spec/codecs/gzip/1.0
+    https://purl.org/zarr/spec/codec/gzip/1.0
 Issue tracking:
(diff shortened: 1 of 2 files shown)


joshmoore:
Anyone want to give this a :thumbsup:? I'd like to get this settled so if anyone else starts implementing V3 there aren't multiple values in the wild.

rouault:
I would also appreciate to see this issue settled in one way or another (pretty much indifferent which solution is adopted) since the new GDAL Zarr driver is ready to be shipped (using the plural form currently): https://github.com/OSGeo/gdal/pull/3896

joshmoore:
Hi @rouault. Thanks for the heads up. I didn't get any :thumbsup: for https://github.com/zarr-developers/zarr-specs/pull/119#issuecomment-866598314 but I also didn't get any :thumbsdown:. 

I'm assuming at this point that that means no vetoes. Let's go with the singular.

PR #135, opened 2022-03-02
Add minimal drafts of a few more dtype extensions

grlee77:
This adds draft extensions for the remaining dtypes supported for Zarr v2 arrays in zarr-python that are not part of the core v3 spec. Merging this would at least give us a URL to point to for these dtypes in https://github.com/zarr-developers/zarr-python/pull/898.

The language here will eventually need updating to be more spec-like and more implementation neutral (e.g. less NumPy/Python specific). What is there now, at least describes the current behavior in zarr-python.

In the last community call @jbms (from from the [tensorstore](https://google.github.io/tensorstore/) team) mentioned that NumPy's fixed width unicode arrays are not very desirable and variable length strings are likely of more interest. There is not a variable length string dtype in NumPy, but variable length strings can be stored as object arrays ([H5Py for example does this](https://docs.h5py.org/en/stable/special.html#variable-length-strings)). I know Pandas 1.0 also introduced their own `StringDType`, but I haven't looked into the specifics of that.

### DIFF
--- a/docs/protocol/extensions.rst
+++ b/docs/protocol/extensions.rst
@@ -11,6 +11,7 @@ Under construction.
    extensions/filters/v1.0
    extensions/complex-dtypes/v1.0
    extensions/datetime-dtypes/v1.0
+   extensions/object-dtypes/v1.0
    extensions/string-dtypes/v1.0
    extensions/struct-dtypes/v1.0
 
--- a/docs/protocol/extensions.rst
+++ b/docs/protocol/extensions.rst
@@ -12,6 +12,7 @@ Under construction.
    extensions/complex-dtypes/v1.0
    extensions/datetime-dtypes/v1.0
    extensions/string-dtypes/v1.0
+   extensions/struct-dtypes/v1.0
 
 
 A number of other features might be included in the core protocol v3, but are currently considered as extensions. 
--- a/docs/protocol/extensions.rst
+++ b/docs/protocol/extensions.rst
@@ -11,6 +11,7 @@ Under construction.
    extensions/filters/v1.0
    extensions/complex-dtypes/v1.0
    extensions/datetime-dtypes/v1.0
+   extensions/string-dtypes/v1.0
 
 
 A number of other features might be included in the core protocol v3, but are currently considered as extensions. 
--- a/docs/protocol/extensions/object-dtypes/v1.0.rst
+++ b/docs/protocol/extensions/object-dtypes/v1.0.rst
@@ -1,0 +1,105 @@
+===================================
+ String data types (version 1.0)
+===================================
+-----------------------------
+ Editor's draft 2 March 2022
+-----------------------------
+
+Specification URI:
+    http://purl.org/zarr/spec/protocol/extensions/object-dtypes/1.0
+Issue tracking:
+    `GitHub issues <https://github.com/zarr-developers/zarr-specs/labels/object-dtypes-v1.0>`_
+Suggest an edit for this spec:
+    `GitHub editor <https://github.com/zarr-developers/zarr-specs/blob/core-protocol-v3.0-dev/docs/protocol/extension/object-dtypes/v1.0.rst>`_
+
+Copyright 2022 `Zarr core development
+team <https://github.com/orgs/zarr-developers/teams/core-devs>`_ (@@TODO
+list institutions?). This work is licensed under a `Creative Commons
+Attribution 3.0 Unported
+License <https://creativecommons.org/licenses/by/3.0/>`_.
+
+----
+
+
+Abstract
+========
+
+This specification is a Zarr protocol extension defining a data type where each
+element is an arbitrary Python object.
+
+
+Status of this document
+=======================
+
+This document is a **Work in Progress**. It may be updated, replaced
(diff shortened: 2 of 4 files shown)


docs/protocol/extensions/object-dtypes/v1.0.rst
```diff
@@ -0,0 +1,105 @@
...
+Object data types
+=================
+NumPy's object arrays are arrays where each element is an arbitrary Python
+object. The array elements correspond to the
+`numpy.object_ <https://numpy.org/doc/1.22/reference/arrays.scalars.html#numpy.object_>`
+type which has character code `'O'`. A common concrete use case for this type
+is to have an array where each element is another array (and each array can
+have a different length). Another use case is to store an array of variable
+length strings. It is important to note that such an array actually just stores the references to the Python objects and not the objects themselves. Accessing
+an element of the array returns the Python object it refers to.
```
jbms, line 72: This should specify how this data type is actually encoded in the store --- presumably via pickling.

In general this sort of encoding issue might fall under the domain of "filters", which are not yet part of zarr v3 but might be important to add.

In general I have a number of concerns about this data type:
- The pickle format by its nature is inherently tied to Python and cannot be supported by non-Python implementations.
- By the nature of the format, the Python unpickle operation can lead to arbitrary code execution and therefore requires that the source data be fully trusted.  I think if it is supported it is likely to be used out of convenience even in situations that don't call for that level of trust.  In particular, if it is available by default in zarr-python then opening any untrusted zarr array becomes a potential security vulnerability.
- In cases where the intent is not to store an arbitrary Python object, but e.g. just a variable length string or json value, just using the "object" type leaves the actual type unspecified.

I think it would be much better instead to decouple the zarr v3 data types from numpy data types --- and define as explicit zarr v3 data types:
- variable length unicode and byte strings
- json values

I think it would be best to avoid supporting a general Python object type at all.

docs/protocol/extensions/string-dtypes/v1.0.rst
```diff
@@ -0,0 +1,136 @@
...
+====================
+
+Two extension data types are defined to represent zero-terminated bytestrings as
+well as fixed-length 32-bit unicode arrays.
+
+Fixed length byte strings (zero-terminated)
+-------------------------------------------
+
+These are fixed width strings corresponding to NumPy dtypes with `kind` 'S'.
+For backward compatibility with Python 2's ``str`` these are zero-terminated
```
jbms, line 75: Zero-terminated strings normally refers to the C null-terminated string convention where the length is not explicitly indicated and instead the end of the string is indicated by a `'\0'` value.  I would call this zero-padded rather than zero-terminated, since there is no zero terminator if the string is exactly equal to the specified length.

docs/protocol/extensions/struct-dtypes/v1.0.rst
```diff
@@ -0,0 +1,127 @@
...
+==================================
+
+.. list-table:: Data types
+   :header-rows: 1
+
+   * - Identifier
+     - Numerical type
+     - Size (no. bytes)
+     - Byte order
+   * - list of (<name>, <type>) tuples
```
jbms, line 99: NumPy (and zarr v2) supports a third variant, where you have `(name, type, shape)`.  The shape specifies the dimensions of an "inner" array of elements.

docs/protocol/extensions/object-dtypes/v1.0.rst
jbms: Note: These variable length strings and json values when stored in a numpy array would have to be stored using the numpy object data type.  But that does not need to match the data type of the zarr array.

docs/protocol/extensions/string-dtypes/v1.0.rst
grlee77: Yeah, I was a bit confused regarding this as the [NumPy dtype docs](https://numpy.org/doc/stable/reference/arrays.dtypes.html) refer to this type as "zero-terminated bytes".

docs/protocol/extensions/struct-dtypes/v1.0.rst
grlee77: Yeah, I had forgotten to list that one and will update it here to describe an optional [, shape]. I think we do test it already.

docs/protocol/extensions.rst
```diff
@@ -11,6 +11,9 @@ Under construction.
    extensions/filters/v1.0
    extensions/complex-dtypes/v1.0
    extensions/datetime-dtypes/v1.0
+   extensions/object-dtypes/v1.0
```
joshmoore, line 14: Mostly as a side note as we work out the process, we may want to lower these numbers depending in our confidence and the intended interpretation.

joshmoore (commented):
Great to have these discussions started, @grlee77! Also :+1: for @jbms' comments, but I could also see getting this into the dev branch before they are resolved as with https://github.com/zarr-developers/zarr-python/pull/898 so that we have more people working from the same starting point.

joshmoore:
Wonderfully, this has no conflicts despite the recent updates! I'm merging and will leave the comments above unresolved for capturing as the editing of v3 moves forward. (If anyone feels moved to create issues, please just cross-link)

PR #142, opened 2022-05-09
Structure spec pages and clarify extension points

jstriebel:
This PR changes the spec in the following ways:
* Restructures the spec pages to more clearly distinguish between the zarr core protocol, protocol extensions, stores and codecs, clarifying the hierarchy and relationships of those entities.
* Changes storage transformers to extension points similar to the others. (Not the ability to have transformers is an extension, but the transformers themselves).
* Lists the different extension points in the `Protocol extensions` section and describes if those are required to read data.
* Groups the extension pages by their extension points.
* Adds a `Stability Policy` section. I assumed the intention is to use sth SemVer-like. If not, I'd still find it helpful to clarify the versioning policy.
* Turns metadata keys into sub-headings to be able to reference them easily.

Please note that the only changes to the actual metadata and protocols are the ones turning storage-transformers to extension-points. Others are just intended to be clarifications or related to the format and structure of the specification itself.

### Motivation
While interacting with the specification to draft the sharding proposal I found it hard to get an overview of the extension points and their different compatibility strategies and behaviors. I tried to improve this here.

### Further steps
The page restructuring might also change the purl URLs. I didn't adapt any of them so far as I wanted to first see if this structure is actually useful to others, too.

If any of those points needs longer discussions I'm happy to separate them from this PR, so other more agreeable points might already be incorporated.

cc @joshmoore @MSanKeys963 @alimanfoo @rabernat @jakirkham @ryan-williams @jbms @normanrz 

PS: I used to the following steps to get auto-updating rendered docs:
```
## optionally setup an venv
# python3 -m venv .venv
# . .venv/bin/activate
pip install -r docs/requirements.txt
pip install sphinx-autobuild
sphinx-autobuild -a docs docs/_build/html
```

Overall structure:


docs/core/v3.0.rst
```diff
@@ -1042,11 +1079,14 @@ The following names are optional:
     compressor is used.
 
 ``storage_transformers``
+^^^^^^^^^^^^^^^^^^^^^^^^
 
     Specifies a stack of `storage transformers`_. Each value in the list must
-    be an object containing the name ``storage_transformer`` whose value
-    is a URI that identifies a storage transformer and dereferences to a
-    human-readable representation of the codec specification. The
+    be an object containing the names ``extension`` and ``type``.
```
jbms, line 1085: I like the use of `type` rather than `storage_transformers` as the identifier key.

However, what is the reason to have both `extension` and `type`?  Is the idea that a single "extension" may define more than one type?

I think it would be simpler to just have a single key, `type`, which is the URI.  If a single extension defines more than one type, the URI could have a fragment identifier, like: `https://whatever/path#some-type`

jbms:
As an editorial comment not specifically related to this PR, I find the use of the term "protocol" rather confusing.  I think it would be better to just use the term "zarr specification" or "zarr format".

"Protocol" is normally used to refer to a sequence of steps / a procedure to follow, especially for communication between multiple programs/entities.

While the current version of the v3 spec does describe certain basic operations under the "Stores" and "Storage protocol" headings, in my view it is the format specification itself that is of primary importance, and these descriptions are more like suggestions for implementers.  In fact I think it may be better to remove most of those descriptions.  However, if there are some operations that require a particularly complex implementation then it could make sense to provide some implementation guidance along with a rationale.

docs/core/v3.0.rst
jstriebel: The reason was to use the same structure `data_type`, `chunk_grid` and `chunk_memory_layout` already have. I think using the same structure across different metadata objects helps for consistency, but I agree that there might be a more elegant solution. What about just using `extension` as the URI key, and having `format` back in the `configuration` object?

jstriebel:
> As an editorial comment not specifically related to this PR, I find the use of the term "protocol" rather confusing. I think it would be better to just use the term "zarr specification" or "zarr format".
> 
> "Protocol" is normally used to refer to a sequence of steps / a procedure to follow, especially for communication between multiple programs/entities.

I was confused by this as well, core zarr specification or sth similar sounds good to me. I'm also wondering what exactly is still part of the v3.0 (core) spec. In my current understanding extensions, codecs and storages are not, that's why I changed the structure this way.

> While the current version of the v3 spec does describe certain basic operations under the "Stores" and "Storage protocol" headings, in my view it is the format specification itself that is of primary importance, and these descriptions are more like suggestions for implementers. In fact I think it may be better to remove most of those descriptions. However, if there are some operations that require a particularly complex implementation then it could make sense to provide some implementation guidance along with a rationale.

I agree that there are multiple things currently mixed in the document:
Explaining the terminology and naming conventions, describing the files and metadata layout, and describing the storage protocol (interactions with the storage). I find the storage protocol very useful to describe possible interactions of stores as well as storage transformers. In my understanding this is mainly establishing vocabulary for the interactions, and provides a non-normative guideline for implementations.

---------------

Both parts seem good proposals for further improvements, but I'd rather keep them separate from this PR if you agree, @jbms. I'd be happy to interact on those on a follow-up PR :+1:

docs/core/v3.0.rst
jbms: Thanks for the clarification, I missed the fact the extension/type structure was already being used.  Maybe for uniformity all of these extension points should just be identified by a url unconditionally even if the "type" is defined by the core specification.

E.g.

```json
{"chunk_layout": {"type": "https://zarr.dev/...#regular_grid", "configuration": ...},
 "data_type": {"type": "https://zarr.dev/...#float32"},
 "compression": {"type": "https://zarr.dev/...#zstd"}
}
```

I think putting the extension urls in a separate "extensions" property is more complicated and also would create the possibility of a conflict if the user wishes to use at the same time two extensions that happen to use the same "type" value.

MSanKeys963:
> As an editorial comment not specifically related to this PR, I find the use of the term "protocol" rather confusing. I think it would be better to just use the term "zarr specification" or "zarr format".
> 
> "Protocol" is normally used to refer to a sequence of steps / a procedure to follow, especially for communication between multiple programs/entities.

I agree with this and I feel if we use 'specification' or 'format' it'd make things simple for us. Do you think you can open a PR for this @jbms? I'd really appreciate it.

docs/core/v3.0.rst
```diff
@@ -137,6 +138,29 @@ provide a framework for community-defined extensions, which can be
 developed and published independently without requiring centralised
 coordination of all specifications.
 
+See `protocol extensions`_ below.
+
+Stability Policy
```
jbms, line 143: This section is somewhat redundant and also inconsistent with the description regarding "zarr_format".  In particular `zarr_format` claims both backwards and forwards compatibility for a given major version number.

What do you have in mind as far as what type of changes would result in a major, minor, or patch number change?

joshmoore:
@jakirkham: any chance of your RTD magic working here?

docs/core/v3.0.rst
```diff
@@ -137,6 +138,29 @@ provide a framework for community-defined extensions, which can be
...
+----------------
+
+This core protocol specification adheres to adapted Semantic Versioning regarding
+read and write compatibility using the ``MAJOR.MINOR[.PATCH]?`` version
+number format, where the patch number is optional and implicitly ``0``.
+A zarr implementation provides the read and write API by implementing
+this specification, the datasets are considered consumers in this case.
+
+Notably, this excludes extensions, codecs and stores from the compatibility
```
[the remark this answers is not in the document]
joshmoore: Isn't the rest of the document thought trying to elevate stores to something more like a proper extension?

rabernat:
> More specifically, interested parties would merge changes into this branch with some pull request review but without the full discussion of the ZEP process, and then once we think the branch is ready present it as ZEP 1 for full feedback.

:+1:

jbms:
I think merging this PR is fine, though I do think we should resolve the "stability" section.

docs/core/v3.0.rst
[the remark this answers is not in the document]
jbms: The spec defines some operations that stores should support, but ultimately it is up to each implementation to decide how stores are implemented and what interfaces they may use.

Individual implementations could certainly document what stores they support and any relevant stability guarantees.

Currently there is nothing in the spec itself as far as specifying which store to use for a zarr root, e.g. a path to a local directory, an HTTP URL, in-memory storage, path to zip file, etc.  If we standardized a URL syntax or JSON representation for specifying where a zarr root/array is stored, then it would make sense to talk about stability guarantees of that URL/JSON representation.

I suppose this spec could describe mappings from the "store" abstraction to concrete storage mechanisms, like local files, zip file, GCS/S3, etc.  However those mappings would be quite trivial so I'm not sure it is very useful to standardize that without also standardizing some URL syntax/JSON representation for specifying the store.

docs/core/v3.0.rst
jstriebel: I'll leave it as-is for now, let's discuss this separately.

docs/core/v3.0.rst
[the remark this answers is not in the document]
jstriebel: I assume stores are described in the core specification via the storage interface, but the actual supported storage backends might still have ambiguities how the key-layout is mapped to a specific backend. This would need to be described separately, and is not part of the core spec IMO. Therefore I'd rather list stores explicitly here as well, an implementation following v3.0 doesn't necessarily say anything about the store compatibility (e.g. if it can read from zip-stores).

docs/core/v3.0.rst
jstriebel: Good catch, I didn't re-check the `zarr_format` contents. I adapted the stability policy now to conform with the `zarr_format` part (no actual semver, but full compatibility across the same major version). There's still two sentences of duplication, not sure if that's fine or too redundant. What do you think about the new version @jbms? If it doesn't provide benefits I'm also not opposed to removing the section again.

docs/core/v3.0.rst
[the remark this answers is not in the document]
joshmoore: > ultimately it is up to each implementation to decide how stores are implemented and what interfaces they may use.

Yeah, I'm not sure about the interfaces of the stores either though the "feature sets" that are listed do tend to be pretty relevant to the rest of the spec. But I think there are also storage issues which are pretty important at this level. For example, `zip my.zarr my.zip` doesn't just work. Zip storage needs to be specified. I think that will be true to some degree for other stores.

> Individual implementations could certainly document what stores they support and any relevant stability guarantees.

Agreed. And I think that's another key role here is to say whether given storage types are MUST, MAY or SHOULD. 

> If we standardized a URL syntax or JSON representation for specifying where a zarr root/array is stored

:+1: which also reminds me that I think of consolidated metadata as a metadata-storage-type.

Issue #133, opened 2022-02-16
Specifying how `fill_value` is handled (if unspecified)


jakirkham:
The Zarr v2 spec [leaves undefined `fill_value`s as ambiguous]( https://zarr.readthedocs.io/en/stable/spec/v2.html?highlight=fill_value#chunks ):

> If the “fill_value” field is null then the contents of the chunk are undefined.

However if a user writes to a portion of a chunk with different implementations, each implementation will now potentially have different chunks depending on how they handle the `fill_value`.

This can also cause confusion in other contexts ( for example https://github.com/zarr-developers/zarr-python/pull/966 ).

Also this gets mentioned in James' overview in issue ( https://github.com/zarr-developers/zarr-specs/issues/53 ).

Of course there are some advantages of leaving this unspecified. Namely one can use uninitialized memory to allocate each chunk for writing into, which would be a bit faster. Also this can more easily handle new or complicated types where an appropriate default `fill_value` is not entirely obvious.

That said, given some of the issues above, wonder if we should take a different tack and specify `fill_value` for types in v3. Thoughts? 🙂

d-v-b:
What about defining the default fill value for a datatype to be the instance of that type with a binary representation of all 0s? Are there any types where this would be counter-intuitive or bad somehow?

jakirkham:
What does that mean for `bytes`, `str`, `object`, etc.?

d-v-b:
I don't know, I never work with those types :) But this "all 0s in its binary representation" idea doesn't seem to make any sense for variable-length types, so it's probably a non-starter. Another option would be to just require a `fill_value`. Explicit is better than implicit and all that.

jbms:
I agree --- fill_value should be required.

There are a few meanings you could potentially assign to an unspecified fill_value:

1. Same as specifying zero or other default value for the type, like empty string.
2. Read returns uninitialized memory --- this is a potential security vulnerability because it may leak information like encryption keys, credentials, etc.
3. Reading a missing chunk returns an error.  However this leads to inconsistent behavior for partially-written chunks, since there is no way to indicate that it is partially written.

Since 1 is the only reasonable option I don't see any advantage in allowing an unspecified fill value.

dopplershift:
[...]

PR #145, opened 2022-06-01
Require `fill_value` to be defined

jakirkham:
Fixes https://github.com/zarr-developers/zarr-specs/issues/133

### DIFF
--- a/docs/core/v3.0.rst
+++ b/docs/core/v3.0.rst
@@ -1033,7 +1033,9 @@ following mandatory names:
 
     For any data type, the ``fill_value`` is required. The literal
     ``null`` is not permitted. The fill value needs to be defined
-    so that the data is independent of implementation details.
+    so that the data is independent of implementation details. Internally
+    implementations may provide a default ``fill_value``, but that must
+    be converted to a fixed value in the stored metadata.
 
     If the ``data_type`` of an array is defined in a ``data_type`` extension,
     then said extension is responsible for interpreting the value of
--- a/docs/core/v3.0.rst
+++ b/docs/core/v3.0.rst
@@ -1031,10 +1031,9 @@ following mandatory names:
     value must be a number with no fraction or exponent part and must
     be within the range of the data type.
 
-    For any data type, if the ``fill_value`` is the literal ``null``
-    then the fill value is undefined and the implementation may use
-    any arbitrary value that is consistent with the data type as the
-    fill value.
+    For any data type, the ``fill_value`` is required. The literal
+    ``null`` is not permitted. The fill value needs to be defined
+    so that the data is independent of implementation details.
 
     If the ``data_type`` of an array is defined in a ``data_type`` extension,
     then said extension is responsible for interpreting the value of


jbms:
This looks good to me.

docs/core/v3.0.rst
```diff
@@ -1031,10 +1031,11 @@ following mandatory names:
     value must be a number with no fraction or exponent part and must
     be within the range of the data type.
 
-    For any data type, if the ``fill_value`` is the literal ``null``
-    then the fill value is undefined and the implementation may use
-    any arbitrary value that is consistent with the data type as the
-    fill value.
+    For any data type, the ``fill_value`` is required. The literal
+    ``null`` is not permitted. The fill value needs to be defined
```
d-v-b, line 1035: wouldn't the literal `null` be a permitted fill value for an array of objects? Maybe just remove the sentence which bars the literal null.

jbms: I agree that it doesn't make sense to specifically exclude `null`.

jakirkham: In Zarr Python, arrays of objects already need an object codec and go through encoding before writing. If we are already handling encoding with fill values (thinking about `base64` discussion), we could expect that happens here (even for `None`).

That said, the original spec didn't cover object types. We may need to do work on [the object type proposal]( https://github.com/zarr-developers/zarr-specs/blob/6b66d484cbf985221dc0badb003b2b3937214d5a/docs/extensions/data-types/object/v1.0.rst#editors-draft-2-march-2022 ) to cover this (cc @jstriebel)

jbms: Agreed that object data types require further discussion (see #146).  But since the allowed values for fill_value are already specified I don't think it is necessary to specifically prohibit null.  For a hypothetical json data type, it could make sense to encode the fill value as a literal json value and in that case null would be valid.

meggart: I agree that it should be unnecessary for exclude `null` since it is already implicitly excluded by the previous sentences. However, given that `null` fill_values are abundant in v2 I think it is worth mentioning here. Maybe we can modify the sentence to something like: This means that in contrast to v2 of this spec ``null`` is not permitted as a fill_value for numeric data types.

jakirkham: The reason to suggest banning `null` is to handle the issue we encountered when trying to skip writing empty chunks. Namely using `null` for `fill_value` is underspecified. One alternative (if we want to solve the write optimization problem) might be to better specify what `null` means

jbms:
Per discussion at the last meeting, I believe there was consensus that this should be merged.  I think the explicit ban on `null` should be removed, since nothing in the spec allows `null` anyway, and for a data type added in the future, `null` could be a valid fill value.  But in any case this PR, as is, is an improvement.

docs/core/v3.0.rst
jstriebel: I agree that it's easier to allow null later if needed, and that it's mostly emphasizing that it's not somehow optional, not even by putting in `null`. If `null` becomes a viable value for some codecs, let's add it then.

jstriebel (approved):
LGTM :+1:

To move forward with the spec I'm merging this. I think everyone agreed to make the `fill_value` required, but feel free to open an issue or PR if that's not the case, or the note about `null` values should be changed. Merging this doesn't mean it's set in stone, but hopefully enables more progress and refinements.

PR #146, opened 2022-06-01
Remove the object data type extension.

jbms:
While this extension provides compatibility with zarr v2, and there
are certainly applications where it makes sense to store arbitrary
Python objects, the current extension has a number of issues:

- As it is currently specified, it describes the in-memory Python
  object array representation, but surely it is not intended that the
  zarr array would actually store Python object memory addresses.

- No encoding is specified (e.g. pickle or json).

- Pickle is a problematic since it is Python-specific and inherently
  unsafe to use to read untrusted data.


jakirkham:
cc @jstriebel @joshmoore

jstriebel:
LGTM, I always assumed that's a "works for python only":tm: extension. In this case there's no point to put it into the spec. If this is still supported in zarr-python (which I'd find valid, even if it's not part of the spec), we should probably reconsider the name/shorthand, but that's a different issue.

jbms:
There are some variants of an "object" type that don't have to be Python-only, like a `json` object type using json, cbor, or msgpack.

Additionally, in principle there could be multiple zarr implementations that support Python, e.g. tensorstore could support Python objects with various codecs, so even if it is Python-specific there could be some benefit in having a specification.

However, I would caution against zarr-python supporting pickle at all in the way it does for zarr v2.  Currently, with zarr v2, if you open an array from an untrusted source, it opens you up to an attacker being able to execute arbitrary code by specifying the pickle codec.  Instead, it is important that the pickle codec require an explicit opt-in by the user when opening the array, e.g. `zarr.open(..., unsafe_codecs=True)`.

jakirkham:
Just a note on context/history, "works with Python" means users need to supply an `object_codec` explicitly (`Pickle` being one of many options). Interestingly the PR that added `Pickle` as a codec also added `MsgPack` ( https://github.com/zarr-developers/numcodecs/pull/5 ). A number of other alternatives also exist `JSON`, `Categorize`, `VLen*`, etc. and were also mentioned in a notebook when adding `object_codec` ( https://github.com/zarr-developers/zarr-python/pull/212 ). So folks have been cognizant about the issues around Pickle and tried to provide reasonable alternatives based on user needs.

Think for the most part when this has come up, the interest has largely been around handling of strings, which can be a bit bumpy in Python. It may be worth opening an issue to discuss strings handling separately.

Other issues might be ragged or other array types. Though those probably deserve their own (use case motivated) discussions to ensure we handle them effectively.

jakirkham:
Thanks Jeremy for the PR and Jonathan for the review! 🙏

PR #151, opened 2022-08-04
sharding spec: minor updates and reflow

jstriebel:
This PR contains minor changes in the sharding specification, some grammar corrections and a reflow of the document.


joshmoore (approved):
I'm assuming this is ok, but in the future I'd vote for separate reflow and update commits for reviewability.

PR #153, opened 2022-08-05
Support a list of `codecs` in place of a single `compressor` field

jbms:
Per what seemed to be the consensus reached at the zarr meeting on 2022-07-27, this replaces the `compressor` field in the array metadata with a `codecs` field that can specify a list of codecs.

### DIFF
--- a/docs/core/v3.0.rst
+++ b/docs/core/v3.0.rst
@@ -361,29 +361,41 @@ conceptual model underpinning the Zarr format.
     ordering of elements_, but extensions may define other
     memory layouts.
 
-.. _compressor:
-.. _compressors:
-
-*Compressor*
-
-    An array_ may be associated with a compressor, which is a codec_
-    that transforms the binary representation of a chunk_ in some way,
-    usually to reduce data size.
-
 .. _codec:
 .. _codecs:
 
 *Codec*
 
-    A codec is a pair of algorithms which transform binary data in
-    some way and are used to encode and decode chunks_. This
-    specification defines a codec interface which comprises
-    a pair of operations, one to perform the transformation (encode),
-    the other to reverse the transformation (decode). This
-    specification only considers the case where a codec is used as a
-    compressor_, but extensions may extend the chunk encoding
-    process, for example, to add support for one or more filter codecs
-    to be applied prior to compression.
+    An array_ may be associated with a list of *codecs*.  Each codec specifies a
+    bidirectional transform (an *encode* transform and a *decode* transform).
+
+    Each codec has an *encoded representation* and a *decoded representation*;
+    each of these two representations are defined to be either:
+
+    - a multi-dimensional array of some shape and data type, or
+    - a byte string.
+
+    Logically, a codec ``c`` must define three properties:
+
+    - ``c.compute_encoded_representation_type(decoded_representation_type)``, a
+      procedure that determines the encoded representation based on the decoded
+      representation and any codec parameters.  In the case of a decoded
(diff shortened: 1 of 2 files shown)


martindurant:
I am very much in favour of this idea

rabernat:
This makes sense to me too.

martindurant:
There is an implicit assumption here, that if the last codec in an encoding chain still produces an array as opposed to bytes (or the list is empty), then the output to storage is the bytes (memory buffer) of that array. That's like the "null" or "dummy" codec mentioned elsewhere. I don't think we need that codes if we are explicit about it here.

jbms:
> There is an implicit assumption here, that if the last codec in an encoding chain still produces an array as opposed to bytes (or the list is empty), then the output to storage is the bytes (memory buffer) of that array. That's like the "null" or "dummy" codec mentioned elsewhere. I don't think we need that codes if we are explicit about it here.

See here: https://github.com/zarr-developers/zarr-specs/pull/153/files#diff-90d755c4d7549e4ab133911db1e21cda80aeba463eb25c00a76501c2ab89a5d2R836

martindurant:
Perfect, thanks @jbms

jstriebel:
LGTM as well. Is there missing anything for this to be merged? cc @joshmoore @MSanKeys963 @alimanfoo @jakirkham We can cherry-pick the change onto the ZEP1-review branch then as well.

docs/core/v3.0.rst
```diff
@@ -757,21 +769,104 @@ elements in the order (0, 0), (1, 0), (2, 0), ..., (`dy` - 3, `dx` -
...
+
+1. The chunk array ``A`` (with a shape equal to the chunk shape, and data type
+   equal to the zarr array data type) is logically transformed into the initial
+   *encoded chunk* ``EC[0]`` of the type specified by
+   ``decoded_representation[0]`` according to the ``chunk_memory_layout`` as
+   follows:
+   
+   - If ``chunk_memory_layout`` is equal to ``"C"``, ``EC[0]`` equals ``A`` (no
+     transformation).
+   - If ``chunk_memory_layout`` is equal to ``"F"``, the dimension order is reversed.
```
jstriebel, line 829: What does `the dimension order is reversed` mean here exactly? Would this need to change the binary format or just interpret it differently?
In the end, does this drop F-ordered data in the binary representation? I think that would be incovenient, since this means current F-ordered data cannot be read, or at least not with the dimension-order that was intended for the dataset.

jbms: No, this doesn't drop support for F order.   The intent is that we describe the behavior of codecs in terms of logical arrays without concern for their memory layout.   Otherwise, I think it would be much more complicated to have to additionally specify for each codec how it affects the memory layout of the array, in addition to its logical value.  Since I have defined that when converting between a logical array and a byte sequence, the elements are always encoded in C order, to achieve F order we must transpose the dimensions first.  Actually there was some discussion previously of eliminating "chunk_memory_layout" and instead specifying it via a "transpose" codec; if that were done, we would not need this special case in the description.

jstriebel: Ah, I see, the transpose codec would work similarly to the endianness codec in the other PR #155 then. I personally like the `chunk_memory_layout` field more, but the codec seems fine as well.

jstriebel (approved):
LGTM as well :+1:

joshmoore:
Merging per all the :+1:s. One thing this does make me aware of though, is that we should probably start maintaining a list of explicit changes now before the list becomes too long to be applied to the existing V3 implementations.

PR #155, opened 2022-08-10
Change data type names and change endianness to be handled by a codec

jbms:
There seemed to be consensus to rename data types to more commonly-used names, rather than the NumPy type string-derived names.

There also seemed to be potential agreement to handle endianness via a codec rather than as part of the data type, as long as there is a migration path for zarr v2 data that does not require any chunk re-encoding.

### DIFF
--- a/docs/codecs.rst
+++ b/docs/codecs.rst
@@ -178,6 +178,8 @@ header. The format of the encoded buffer is defined in [BLOSC]_. The
 reference implementation is provided by the `c-blosc library
 <https://github.com/Blosc/c-blosc>`_.
 
+.. _endian-codec:
+
 Endian
 ------
 
--- a/docs/codecs.rst
+++ b/docs/codecs.rst
@@ -206,7 +206,9 @@ type is encoded as a 4-byte big endian two's complement integer, and the
 
    Single the default binary representation of all data types is little endian,
    specifying this codec with ``endian`` equal to ``"little"`` is equivalent to
-   omitting this codec.
+   omitting this codec, because if this codec is omitted, the default binary
+   representation of the data type, which is always little endian, is used
+   instead.
 
 Deprecated codecs
 =================
--- a/docs/codecs.rst
+++ b/docs/codecs.rst
@@ -178,6 +178,35 @@ header. The format of the encoded buffer is defined in [BLOSC]_. The
 reference implementation is provided by the `c-blosc library
 <https://github.com/Blosc/c-blosc>`_.
 
+Endian
+------
+
+Codec URI:
+    https://purl.org/zarr/spec/codec/endian
+
+Encodes array elements using the specified endianness.
+
+Configuration parameters
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+endian:
+    Required.  A string equal to either ``"big"`` or ``"little"``.
+
+Format and algorithm
+~~~~~~~~~~~~~~~~~~~~
+
+Each element of the array is encoded using the specified endian variant of its
+default binary representation.  Array elements are encoded in lexicographical
+order.  For example, with ``endian`` specified as ``big``, the ``int32`` data
+type is encoded as a 4-byte big endian two's complement integer, and the
+``complex128`` data type is encoded as two consecutive 8-byte big endian IEEE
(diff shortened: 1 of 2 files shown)


jakirkham:
cc @QuLogic (in case you have thoughts here)

martindurant:
I believe all FITS integer values are big-endian, for reference.

docs/codecs.rst
```diff
@@ -178,6 +178,35 @@ header. The format of the encoded buffer is defined in [BLOSC]_. The
...
+order.  For example, with ``endian`` specified as ``big``, the ``int32`` data
+type is encoded as a 4-byte big endian two's complement integer, and the
+``complex128`` data type is encoded as two consecutive 8-byte big endian IEEE
+754 binary64 values.
+
+.. note::
+
+   Single the default binary representation of all data types is little endian,
+   specifying this codec with ``endian`` equal to ``"little"`` is equivalent to
+   omitting this codec.
```
jstriebel, line 209: If I understand this correctly, an implementation could also decide to return a big-endian representation, which reverts this logic, right?

docs/core/v3.0.rst
```diff
@@ -488,101 +486,48 @@ Core data types
```
jstriebel, line 486: IMO a note about the little-endian defaults which can be changed via the `Endian` codec would be great in the `Data types` section above.

docs/codecs.rst
jbms: I would say that the in-memory representation used by a zarr implementation is out-of-scope for the specification.

But I should clarify that "equivalent to omitting this codec" is just because the default conversion from logical array to byte sequence uses the default binary representation, which is always little endian.  However, "equivalent to omitting this codec" is not the same as "no byte order conversion from the in-memory representation is performed"; indeed on a big endian machine the in-memory representation would likely be big endian and therefore if the endian codec is not specified with big endian, then a byte order conversion will be needed to convert to the binary representation.

jstriebel: Great, thanks :+1: From my side this LGTM, when those clarifications are added.

jbms: Fixed.

docs/core/v3.0.rst
jbms: Done.

docs/core/v3.0.rst
```diff
@@ -591,6 +536,11 @@ should be understood as fall-back types of respectively 1, 2, and 3 byte length.
 Zarr v3 is limited to type sizes that are a multiple of 8 bits but may support
 other type sizes in later versions of this specification.
 
+.. note::
+
+   While the default binary representation is little endian, the :ref:`endian
+   codec<endian-codec>` may specified to use big endian encoding instead.
```
jstriebel, line 542:
```suggestion
   codec<endian-codec>` may specify to use big endian encoding instead.
```

jstriebel:
LGTM :+1:

docs/core/v3.0.rst
jbms: Done.

meggart (approved):

jstriebel (approved):

joshmoore:
Merging based on current approvals.

docs/codecs.rst
```diff
@@ -178,6 +178,39 @@ header. The format of the encoded buffer is defined in [BLOSC]_. The
...
+Each element of the array is encoded using the specified endian variant of its
+default binary representation.  Array elements are encoded in lexicographical
+order.  For example, with ``endian`` specified as ``big``, the ``int32`` data
+type is encoded as a 4-byte big endian two's complement integer, and the
+``complex128`` data type is encoded as two consecutive 8-byte big endian IEEE
+754 binary64 values.
+
+.. note::
+
+   Single the default binary representation of all data types is little endian,
```
QuLogic, line 209: Since the default binary representation of all data types is little endian,

QuLogic (commented):
Just noticed a small typo.

docs/codecs.rst
joshmoore: https://github.com/zarr-developers/zarr-specs/pull/165

PR #149, opened 2022-07-21
ZEP0001 - Core v3.0 spec for review

alimanfoo:
This pull request includes the Zarr core specification v3.0 as proposed in [ZEP0001](https://zarr.dev/zeps/draft/ZEP0001.html).

Reviews of this specification are now invited. If possible, please submit comments by using the normal GitHub code review functionality.

Technical note: Since the ZEP process has been adopted, the normal process for proposing new or modified Zarr specifications is to submit a ZEP and to create an associated pull request to the zarr-specs repo with the proposed specification content. However, work on the v3.0 spec began before the ZEP process was adopted, and a draft of the spec had already been merged to the main branch. To make reviewing this spec easier and most similar to the normal process going forward, this PR has been submitted against a "void" branch with no content.

PR #162, opened 2022-11-02
Add `dimension_names` array metadata field

jbms:


jstriebel:
Fixes #73

docs/core/v3.0.rst
```diff
@@ -1095,6 +1091,18 @@ The following names are optional:
     specification. When the ``storage_transformers`` name is absent no storage
     transformer is used, same for an empty list.
 
+``dimension_names``
+^^^^^^^^^^^^^^^^^^^
+
+    Specifies dimension names, e.g. ``["x", "y", "z"]``.  If specified, must be
+    an array of strings with the same length as ``shape``.  An unnamed dimension
+    is indicated by an empty string, ``""``.  Within a single zarr array, all
```
shoyer, line 1099: Why not use `null` instead for unnamed dimensions?

docs/core/v3.0.rst
```diff
@@ -1095,6 +1091,18 @@ The following names are optional:
...
+    non-empty dimension names must be distinct (no two dimensions may have the
```
shoyer, line 1100: I'm not sure how I feel about the uniqueness restriction. I would be inclined to omit it, and leave it up to applications/extensions to enforce.

jbms: Do you have in mind a use case for duplicate dimension names, or do you know of any software that could make use of that?

shoyer: Duplicate dimension names come up for cases like covariance matrices. In Xarray, we support creating them (I have mixed feelings about this), but you can't do named operations in the case of duplicates.

docs/core/v3.0.rst
shoyer: I would prefer enforcing this at the type level, e.g., "dimension names are any valid string."

Otherwise, I imagine _somebody_ is going to try to set dimension names to an empty string in their application code and wonder why they don't round-trip to Zarr properly.

docs/core/v3.0.rst
jbms: Thanks for that clarification --- I didn't think about that use case.  I guess in xarray the dimension names serve (at least) two different purposes: associating coordinate arrays and allowing dimensions to be referenced by name, and duplicate dimension names still are useful for specifying coordinate arrays.  Does netcdf also allow duplicate dimension names?

In tensorstore dimension names are solely used for allowing dimensions to be referenced by name, and duplicate names are not currently supported.

docs/core/v3.0.rst
jbms: The current proposal will allow any string to round trip ---- the uniqueness constraint just applies to non-empty strings.  Are you proposing that empty strings should be supported and be treated exactly the same as  non-empty strings --- and that unnamed dimensions should be identified separately using null?

shoyer: > Are you proposing that empty strings should be supported and be treated exactly the same as non-empty strings --- and that unnamed dimensions should be identified separately using null?

Yes. Otherwise application code will not be able to make the distinction (if desired) between "unnamed" and "named with an empty string."

docs/core/v3.0.rst
shoyer: Yes, NetCDF does allow for duplicate dimension names.

ivirshup: * Cross referencing some relevant xarray issue: https://github.com/pydata/xarray/issues/3731

Another use case here is storing graphs as adjacency matrices. In AnnData (but also other projects like pytorch-geometric) we have an set of observations (nodes) with features in tables and tensors,  aligned with graphs represented as adjacency matrices. If we want to subset or sort the observations, it's important that these operations are applied to both dimensions of the adjacency matrix.

This use case has been coming up a lot in discussions around AnnData's integration into OME-NGFF, where you'd have a graph connecting neighboring regions.

jbms: Do NetCDF implementations support indexing operations using dimension names to specify dimensions?  If so, how do they handle duplicate dimension names?  Is there any software that actually provides the desired behavior, where if you perform an indexing operation using a named dimension, and there is more than one dimension with the given name, then the indexing operation applies to all of the dimensions?

It seems to me that duplicate dimension names just don't work very well --- if you have duplicate dimension names, then even if there are some indexing operations that should be applied uniformly to all of the identically-named dimensions, there will almost surely be other indexing operations that should be applied to only one or some of the identically-named dimensions --- for example, to look up a non-diagonal entry in the covariance matrix or adjacency matrix.  Since there is no way to uniquely identify such dimensions by name, they have to be specified by dimension index, and you therefore lose the benefit of dimension names.

A better solution might be to keep dimension names unique, but add support for a special syntax to xarray to allow the dimension name not to exactly equal to the coordinate array name.  For example, the coordinate array might be named "x" and the dimension names might be "x:1" and "x:2" (just to toss out an arbitrary syntax that I haven't given much thought to).  xarray could still allow indexing operations on "x" to apply to both dimensions.

jstriebel:
Crosslinking https://github.com/zarr-developers/zarr-specs/pull/149#discussion_r927300522

docs/core/v3.0.rst
shoyer: > Do NetCDF implementations support indexing operations using dimension names to specify dimensions?

The official netCDF-C library does not. Dimension names are just kept around as metadata.

> It seems to me that duplicate dimension names just don't work very well

I agree. Certainly the semantics of Xarray's indexing with labeled dimensions break with duplicates.

That said, I don't think "not working well" in certain data models is a good reason for making the Zarr spec more complex to prohibit certain behavior. There are clearly cases that could justify storing arrays with duplicate dimensions names (e.g., covariance and adjacency matrices), especially if not using APIs that look-up dimensions by name.

Xarray and NetCDF do impose some additional restrictions on dimension names, namely:

- No missing dimension names
- The size of array axes corresponding to a dimension name must be consistent between all arrays in a group

These restrictions make sense for Xarray/NetCDF, but I would not impose them for all uses of `dimension_names` in Zarr. Instead, at most I would add a comment like "Extensions may impose additional restrictions on valid dimension names" with a link to the NetCDF Zarr spec (when complete).

> A better solution might be to keep dimension names unique, but add support for a special syntax to xarray to allow the dimension name not to exactly equal to the coordinate array name.

I agree, this sort of thing could make sense for Xarray.

jbms: Per your comment, I changed the requirement to be a recommendation.

ivirshup: > Is there any software that actually provides the desired behavior, where if you perform an indexing operation using a named dimension

AnnData sorta does this, but it's a very constrained data model where we don't really need to do matching dimensions by name.

------

I think the current language resolves this for me.

rabernat (approved):

jstriebel (dismissed):
LGTM :+1:

jstriebel (dismissed):

jstriebel:
@rabernat I just resolved a merge conflict which dismissed our old reviews, and mine doesn't count anymore. Could you re-approve this one please? Thanks :-)

shoyer:
I don't mean to hold things up with bike-shedding, but did we reach concensus on using `""` vs `null` for missing dimension names?

jbms:
> I don't mean to hold things up with bike-shedding, but did we reach concensus on using `""` vs `null` for missing dimension names?

Is your intention to allow the empty string as a valid dimension name distinct from null, and not meaning an unlabeled dimension?

shoyer:
> > I don't mean to hold things up with bike-shedding, but did we reach concensus on using `""` vs `null` for missing dimension names?
> 
> Is your intention to allow the empty string as a valid dimension name distinct from null, and not meaning an unlabeled dimension?

My thinking was that `null` instead of a string would indicate missing dimension names, with no special meaning for empty strings.

I prefer `null` because it's less ambiguous from a typing perspective -- if somebody writes code that doesn't anticipate partially missing dimension names, it will fail noisily rather than silently.

jstriebel:
@jbms As discussed last week, I changed this here to use null objects for unnamed dimensions, as requested by @shoyer. I believe this is ready to be merged now. My approval does not count anymore since I did some changes, @joshmoore could you approve this one?

rabernat (approved):

rabernat:
I got you @jstriebel! 🚀

Issue #33, opened 2019-05-22
Extension slicing: or what to do when things go wrong


joshmoore:
cF: https://doc.zeroc.com/ice/3.6/client-server-features/slicing-values-and-exceptions

Given this class hierarchy defined for the Ice protocol:

```
module ex {

  // Some base data type
  class ZString {
       Bytes data;
  };

  // A particular interpretation of that type
  class ZUTF8 extends ZString {
  };

  // Something arbitrarily more complex ...
  class ZHTML extends ZUTF8 {
      Language lang;
  };

};
```

a `ZHTML` instance that was serialized would contain an ordered list of its extension specs as:

```
In[1]: print(ZHTML().ice_ids())
Out[1]: ["::ex::ZHTML", "::ex::ZUTF8", "::ex::ZString", "::Ice::Object"]
```

A client which wanted to deserialize this object but did not have access to the ZHTML spec would _slice_ off that interpretation and try the next one. If the only spec that was available was `ZString`, then that instance would be of type `ZString` and the client would use the regular handling for that type.

Types that might benefit from this mechanism:
 * complex numerics (e.g. `extends LongPair` if the spec doesn't support the more flexible `Long[2]`)
 * timestamps (e.g. `extends String`)
 * Perhaps https://github.com/zarr-developers/zarr-specs/issues/23 (e.g. `Chunk[]` of different sizes)

tam203:
So I'm not sure about the best implementation but I like the idea of being able to 'descend' from high objects to lower (or even primitives) when not available. The example of dates going to strings (or int's depending on implementation) is a good one. 

I'm nervous (but equally not sure there is a better option) about how the spec is aiming to be very light/minimal with extensions handling many/most of the more complex work. My concern is each institute/individual uses their own combination of filters/compressors/etc and that data becomes less interpretable as everyone is using different combinations that don't necessarily port to different languages/versions and data will rust and degrade as this myriad of libraries aren't maintained or updated.

I realise the above paragraph is all "grump" and no solutions but I think this issue (if implemented well) could help alleviate some of this. 

👏

jstriebel:
[...]

PR #167, opened 2022-11-18
Core Spec: Allow multiple fallback values for data type

jstriebel:
This PR allows to have a list of fallback data types for extensions, where all but the last may stem from the extension as well. This allows implementations to add only basic support for an extension, if multiple data types are defined. We might even want to relax this and allow any data types from any other extensions. Probably the data type name should be unique across extensions anyways, so this would be easy to change.

Fixes issue #33, please also see the thread there.

### DIFF
--- a/docs/core/v3.0.rst
+++ b/docs/core/v3.0.rst
@@ -1037,12 +1037,14 @@ following mandatory names:
     the specification.  The ``type`` is required and its value is
     defined by the extension. The ``fallback`` is optional
     and, if provided, its value must be one of the data type
-    identifiers defined in this specification. If an implementation
-    does not recognise the extension, but a ``fallback`` is present,
-    then the implementation may proceed using the ``fallback`` value
-    as the data type. For fallback types that do not correspond to base
-    known types, extensions can fallback on a raw number of bytes using
-    the raw type (``r*``).
+    identifiers defined in this specification, or a list of identifiers
+    from the extension or the core specifiaction where the last fallback
+    value should be a data type identifier of this core specification. If
+    an implementation does not recognise the extension or specific data
+    type, but a ``fallback`` is present, then the implementation may proceed
+    using the first known ``fallback`` value as the data type. If no fallback
+    types are given or they do not correspond to base known types, extensions
+    can fallback on a raw number of bytes using the raw type (``r*``).
 
 ``chunk_grid``
 ^^^^^^^^^^^^^^


jbms:
It might be easier to postpone considering this feature until we are also standardizing an extension data type for which it might be helpful to specify a fallback.

I don't think there is any reason to require the fallbacks to be from the same extension or core spec --- we may as well allow arbitrary fallbacks.  In particular, for variable length or otherwise specially encoded values, there might not exist any compatible core data type.

I think, though, that using the fallback should require explicit opt-in from the user --- it is not good in non-interactive use cases for the behavior to possibly change silently when using an older version of the zarr library.

jstriebel:
> It might be easier to postpone considering this feature until we are also standardizing an extension data type for which it might be helpful to specify a fallback.

:+1: Sounds good to me. I guess we should finalize an extension data type to verify the extension mechanism in v3 as well. Simply picking one of the existing drafts would probably be fine: https://zarr-specs.readthedocs.io/en/latest/extensions/data-types.html
Just opened #169 with similar points

> I don't think there is any reason to require the fallbacks to be from the same extension or core spec --- we may as well allow arbitrary fallbacks. In particular, for variable length or otherwise specially encoded values, there might not exist any compatible core data type.

:+1: Sounds good to me, will adapt here.

> I think, though, that using the fallback should require explicit opt-in from the user --- it is not good in non-interactive use cases for the behavior to possibly change silently when using an older version of the zarr library

IMO that's an implementation detail and up to the actual implementation. I'll add a short note, though.

docs/core/v3.0.rst
```diff
@@ -1036,13 +1036,16 @@ following mandatory names:
...
-    as the data type. For fallback types that do not correspond to base
-    known types, extensions can fallback on a raw number of bytes using
-    the raw type (``r*``).
+    and, if provided, its value must be one or a list of the data type
+    identifiers defined in this specification or an extension. If
+    an implementation does not recognise the extension or specific data
+    type, but a ``fallback`` is present, then the implementation may proceed
+    using the first known ``fallback`` value as the data type. If no fallback
+    types are given or they do not correspond to base known types, extensions
+    can fallback on a raw number of bytes using the raw type (``r*``).
```
jbms, line 1045: This says "if no fallback types are given" but then says that a raw type may be used.  Instead, it can simply suggest that for fixed-size data types, if there is no more specific fallback available, then a raw type may be specified as a fallback.

It should also note that the default list of fallbacks is defined by the data type extension, but may be overridden by the user.

docs/core/v3.0.rst
```diff
@@ -1036,13 +1036,16 @@ following mandatory names:
...
+    *Note for implementations*: Silently using a fallback without interactive
```
jbms, line 1046: Instead of "interactive", perhaps say "explicit"

jstriebel:
```suggestion
    *Note for implementations*: Silently using a fallback without explicit
```

docs/core/v3.0.rst
jstriebel, line 1045:
```suggestion
    using the first known ``fallback`` value as the data type. For fixed-sized data types,
    if there is no more specific fallback available, a raw number of bytes using the raw type (``r*``)
    should be given.

    The default list of fallbacks to put into the metadata should by defined by the data type extension,
    but it may be overridden by the user.
```

jbms (approved):

joshmoore (approved):
- :+1: for the multivalue, first win semantics
 - :+1: for the implementation note
 - I do wonder if this is becoming a case for (eventually) introducing more examples into the spec document itself.

joshmoore:
Sorry, @jstriebel. I noticed too late that this does not have a changelog entry. (I assume that's something that should be in each and every PR?) I added a card to the project board as a reminder.

PR #168, opened 2022-11-18
Add changelog to v3 core spec and update frontmatter

jstriebel:
This PR
* adds a changelog for the v3.0 core spec regarding implementation relevant changes. This is supposed to help implementations that track this spec to update with the spec.
* updates the frontmatter across different pages, mostly emphasizing the current status via a warning and updating some links. I also included some feedback about the frontmatter from #149.
* removes the version pinning for sphinx, as the old sphinx version does not work with current (unrestricted) versions of it's dependencies. The output of the current version conforms to the old one.

Issue #141, opened 2022-05-04
Support for inf, nan, binary data in attributes


jbms:
In both v2 and v3 specs, attributes are stored as JSON which means they are subject to the limitations of JSON:

- Infinity and NaN cannot be represented, although currently zarr-python supports these via a non-standard extension to JSON format (directly encoding them as unquoted `Infinity` or `NaN` within the JSON output).  This support may be accidental rather than intentional, though.
- Byte strings (as opposed to Unicode strings) cannot be represented.

The v3 spec suggests the possibility of using a binary encoding like CBOR for metadata which would solve these issues.  However my understanding is that CBOR would not be the default.

jakirkham:
Here's [the current code in Zarr Python]( https://github.com/zarr-developers/zarr-python/blob/fd979fbda3cf52dc79c3b1e49a8da0a75b03078d/zarr/meta.py#L230-L235 ) for handling these values. One thought might be to graduate this to inclusion in the spec as-is. Idk if there are other approaches worth considering.

For more complex objects Base64 encoding has come up as an idea.

jbms:
For `fill_value` the data type is already known so there isn't an issue there.  zarr-python for v2 already uses a different encoding for `fill_value` ---- infinity is encoded as `"Infinity"` rather than `Infinity`.

I think it would be nice if `fill_value` can be expressed in a readable way but that isn't critical.

I intended this issue to be about user-defined attributes, though.  For example, a common use case might be to store the min/max range of an array, or the bucket partitions for a histogram as attributes.

jakirkham:
Ah sorry thought from our conversation earlier this was `fill_value` focused.

That being said, the issue is more or less the same in either case. We are storing values in JSON that may not representable using standard JSON types.

In these cases we do know how we would encode the data to write in a binary file. That same strategy could be leveraged to binary encode those values as if they were going in a chunk then convert to `base64` before storing in JSON as string (possibly with an object indicating the encoding like above).

Understand the concern about human readability. With things like NaN and Infinity, we could quote them if that is sufficient. Zarr Python [should be doing this currently]( https://github.com/zarr-developers/zarr-python/blob/fd979fbda3cf52dc79c3b1e49a8da0a75b03078d/docs/spec/v2.rst#fill-value-encoding ). If it is not, that is a bug. We could also attempt JSON encoding these values in a readable way if possible. Only falling back to `base64` if that is not an option.

jbms:
[...]

PR #171, opened 2022-11-21
Apply more ZEP 1 feedback

jstriebel:
This PR applies review feedback from the ZEP1 review in PR #149. Most changes are clarifications and wording corrections. The only further reaching changes are about `metadata_key_suffix` and `metadata_encoding`, explained below. The changes for those fields are not finalized yet, and should be discussed further in #174, separate from this PR, to unblock the other changes. Sorry about mixing those up. If the related changes do not seem acceptable for the moment, I'd rather fall back to removing them and solve this in a separate issue/PR.

Non editorial changes:
* Moved `metadata_key_suffix` into the `metadata_encoding` metadata field, specifying only the encoding schema there, not the metadata content. Changed `metadata_encoding` to be an extension point.


docs/core/v3.0.rst
```diff
@@ -1160,8 +1158,8 @@ The following members are optional:
     transformer is used, same for an empty list.
 
 
-All other names within the array metadata object are reserved for
-future versions of this specification.
+The array metadata object must not contain any other names.
```
jbms, line 1161: This could be clarified to state that that an implementation MUST fail to open a zarr repo/array with unknown metadata fields; that would allow required extensions to be defined in a more concise way.

docs/core/v3.0.rst
```diff
@@ -932,25 +936,21 @@ containing the following names:
...
-
-    For document using the default JSON encoding and format describe in this document
-    then the value must be ``"https://purl.org/zarr/spec/core/3.0``.
-
 ``metadata_key_suffix``
 ^^^^^^^^^^^^^^^^^^^^^^^
 
-    A string containing a suffix to add to the metadata keys when saving into
-    the store. By default ``".json"``.
+    A string containing a suffix to add to the array and group metadata keys
```
jbms, line 942: What do you see as the advantage of using `metadata_key_suffix` rather than `metadata_encoding` to identify the metadata encoding?  I think it may be clearer to specify `metadata_encoding` and instead eliminate `metadata_key_suffix`, since in principle the same extension could be used for two different encodings, etc. 

However, I actually think we should just eliminate metadata_encoding for now, and can worry about it later if there is a concrete proposal for an alternative metadata encoding that someone wants to add.

docs/core/v3.0.rst
```diff
@@ -689,6 +687,12 @@ arbitrary length in a "negative" direction along any dimension.
    in which case the coordinate of a chunk is the empty tuple, and the chunk key
    will consist of the string ``c``.
 
+.. note:: Chunks at the border of an array always have the full chunk size, even when
+   the array only covers parts of it. For example, having and array with ``shape=(30, 30)``,
```
jbms, line 691: "having an array"

docs/core/v3.0.rst
```diff
@@ -689,6 +687,12 @@ arbitrary length in a "negative" direction along any dimension.
...
+   ``chunks=(16, 16)``, the chunk ``0,1`` would also contain unused values for the indices
+   ``(0-16, 30-31)``. Since reading anything beyond the array shape is undefined behavior,
+   those values may be passed on, be replaced with the fill values or other undefined behavior.
```
jbms, line 694: Also this example is using zarr v2 metadata and Python syntax.  It should instead use zarr v3 metadata and json syntax.

As far as "reading anything beyond the array shape is undefined behavior" --- that seems like an implementation concern more than a spec concern.  E.g. some implementations might return an error, while others might do something else.

But the content of those "unused values" stored in the chunk is not necessarily unimportant --- if you resize an array to have larger bounds, and (as is likely to be the case) do not rewrite existing chunks, then what were previously unused values become valid values.  I think, therefore, that the most natural choice for those "unused values" in the chunk is that they are set to the fill value.

docs/core/v3.0.rst
jstriebel:
```suggestion
   the array only covers parts of it. For example, having an array with ``shape=(30, 30)``,
```

docs/core/v3.0.rst
jstriebel: > that would allow required extensions to be defined in a more concise way.

Not sure what you mean here, what would change regarding required extensions?

Anyways, I'll add this.

docs/core/v3.0.rst
```diff
@@ -1160,8 +1158,8 @@ The following members are optional:
     transformer is used, same for an empty list.
 
 
-All other names within the array metadata object are reserved for
-future versions of this specification.
+The array metadata object must not contain any other names.
+Those are reserved for future versions of this specification.
```
jstriebel, line 1162:
```suggestion
Those are reserved for future versions of this specification.
An implementation must fail top open zarr hierarchies, groups
or arrays with unknown metadata fields.
```

docs/core/v3.0.rst
jstriebel: > Also this example is using zarr v2 metadata and Python syntax. It should instead use zarr v3 metadata and json syntax.

Good point :+1:

> As far as "reading anything beyond the array shape is undefined behavior" --- that seems like an implementation concern more than a spec concern. E.g. some implementations might return an error, while others might do something else.

I think that's exactly what undefined behavior means, that implementations are free to do anything. It's only undefined regarding the spec, and implementations might choose to define it as they want.

> But the content of those "unused values" stored in the chunk is not necessarily unimportant --- if you resize an array to have larger bounds, and (as is likely to be the case) do not rewrite existing chunks, then what were previously unused values become valid values. I think, therefore, that the most natural choice for those "unused values" in the chunk is that they are set to the fill value.

I would rather not put that as a requirement, since it would also require implementations to re-write chunks when resizing an array to smaller bounds, which is not strictly needed and might not be wanted for performance reasons or to be able to restore the previous content when resizing to the previous bound again. Since there are multiple possible valid strategies I'd rather not specify an explicit one in the spec.

docs/core/v3.0.rst
jbms: I think we should just remove both for now.  We can still later add a `metadata_encoding` attribute if we want, rather than adding it under "extensions" --- since any volume using a different `metadata_encoding` will only be usable by implementations that support it, the default behavior of failing when encountering any unknown metadata members will do the right thing.

As far as providing a hint to implementations that the extension may change --- I can't imagine it being very difficult to change that later in any implementation.  In general when writing software I think trying to add additional abstraction layers/indirection in anticipation of *future* extensions that you haven't already fully thought through isn't that helpful --- you are likely as not to guess wrong, and just end up doing extra work and having extra complexity from unused additional indirection/abstraction layers.

docs/core/v3.0.rst
jbms: Agreed that we don't want to require implementations to rewrite chunks when shrinking the bounds --- indeed TensorStore does not rewrite chunks when shrinking the bounds.  When writing initially, though, filling out-of-bounds locations with the fill value seems like an obvious choice, without any real downside.  Because the implementation that initially writes the chunk may not be the same as the implementation that does the resize, I think it is still useful to specify the behavior.

docs/core/v3.0.rst
jstriebel: Also referencing #37, where exactly such an extension point was agreed on. I'd rather keep this in atm and tend to making `metadata_encoding` a regular extension point.

> As far as providing a hint to implementations that the extension may change --- I can't imagine it being very difficult to change that later in any implementation. In general when writing software I think trying to add additional abstraction layers/indirection in anticipation of future extensions that you haven't already fully thought through isn't that helpful --- you are likely as not to guess wrong, and just end up doing extra work and having extra complexity from unused additional indirection/abstraction layers.

Agreed, but since this topic came up multiple times (#37, #141), and the extension interface is still quite generic, I think it's worth a try. The only difference is to make this an explicit extension point for metadata encoding, instead of an arbitrary array extension.

jstriebel: It probably was a mistake to mix this topic with other rather smaller things in this PR. I'll go forward with `metadata_encoding` as an extension for now, but opened up a separate issue to discuss if this should be removed #174

docs/core/v3.0.rst
jstriebel: Added this as a recommendation in 4389e12.

jstriebel:
Waiting with asking for reviews for #168 to be merged since this PR was based on it to have a clean diff.

docs/core/v3.0.rst
```diff
@@ -262,7 +258,7 @@ The following figure illustrates the first part of the terminology:
     Each node in a hierarchy_ has a name, which is a string of
     characters with some additional constraints defined in the section
     on `node names`_ below. Two sibling nodes cannot have the same
-    name. The root node does not have a name.
+    name. The root node does not have a name and is the empty string ``""``.
```
rabernat, line 261: This is inconsistent with what is said just below in line 274:
> The path "/"  identifies the root node.

Do we want to say that either the empty string or "/" is a valid specifier of the root node? Or is this up to the implementation?

docs/core/v3.0.rst
```diff
@@ -271,17 +267,15 @@ The following figure illustrates the first part of the terminology:
...
     followed by the name_ of the node itself. For example, the path
     "/foo/bar" identifies a node named "bar", whose parent is named
     "foo", whose parent is the root of the hierarchy. The path "/"
     identifies the root node.
 
-    A path always starts with ``/``.
-
-    QUESTION: do we want to codify that group ``path`` end, can end, or must end
-    in ``/`` ?
+    A path always starts with ``/`` and cannot end with ``/``,
```
rabernat, line 277: Currently, implementations support specifying the path without the initial slash, e.g. `foo/bar`.

docs/core/v3.0.rst
```diff
@@ -459,6 +452,7 @@ terminology for a use case of reading from an array:
 Node names
 ==========
 
+The root node does not have a name and is the empty string ``""``.
```
rabernat, line 455: See comment above about "/".

docs/core/v3.0.rst
```diff
@@ -936,22 +937,27 @@ containing the following names:
...
-    A string containing a suffix to add to the metadata keys when saving into
-    the store. By default ``".json"``.
+    The ``metadata_encoding`` value is an extension point and may be defined by
+    an extension. In this case the value must be an object containing the
+    required names ``extension``, ``type`` and ``metadata_key_suffix``.
+    ``extension`` must be a URI that identifies the extension and dereferences
+    to a human-readable representation of the specification. ``type`` is a
+    string defined by the extension. The ``metadata_key_suffix`` is a string
+    containing a suffix to add to the array and group metadata keys when saving
+    into the store.
```
rabernat, line 956: Question: is "metadata_key_suffix" allowed to be an empty string?

docs/core/v3.0.rst
```diff
@@ -1162,8 +1172,10 @@ The following members are optional:
     transformer is used, same for an empty list.
 
 
-All other names within the array metadata object are reserved for
-future versions of this specification.
+The array metadata object must not contain any other names.
+Those are reserved for future versions of this specification.
+An implementation must fail top open zarr hierarchies, groups
```
rabernat, line 1177:
```suggestion
An implementation must fail to open zarr hierarchies, groups
```

rabernat (commented):
This is looking good. 

I noted an inconsistency in how we are identifying the root node.

rabernat (commented):
This is looking good. 

I noted an inconsistency in how we are identifying the root node.

docs/core/v3.0.rst
jstriebel: I think this is consistent, since `""` is the *node name* of the root node and `"/"` is the *path* of the root node, just as `"bar"` can be a name of a first level node at the path `"/bar"`. Does this make sense?

docs/core/v3.0.rst
jstriebel: I think this is fine to allow it in an API of an implementation, but when such paths are in the spec or metadata I believe it's good to have a more strict guideline. Should we clarify this or do you think it's fine as-is?

docs/core/v3.0.rst
jstriebel: I don't see any downsides, so why not? This would need to be defined in an extension though. See also #174

jstriebel:
@jbms @rabernat Is there anything else missing IYO? I'd love to merge this soon to avoid conflicts and to have the feedback incorporated into https://zarr-specs.readthedocs.io/en/latest/core/v3.0.html.

rabernat (approved):

jstriebel:
Thanks @rabernat :+1:

PR #180, opened 2022-11-24
use sphinxmark to add draft watermark to all pages

### LINKED ISSUE #170: Potentially add "Draft" watermark
After looking at several of the changes in #168, I realized that it might be "safer" to add a watermark to the pages. Example of such changes are available at https://github.com/bids-standard/bids-specification/pull/383

image

jstriebel:
This is a POC for a draft watermark using the [sphinxmark extension](https://kallimachos.github.io/sphinxmark). Currently it's only possible to mark all pages or none, see https://github.com/kallimachos/sphinxmark/issues/5. Might fix #170.


docs/core/v3.0.rst
```diff
@@ -4,9 +4,9 @@
 ======================================
  Zarr core specification (version 3.0)
 ======================================
--------------------------------
- Editor's draft 25 May 2022
--------------------------------
```
jstriebel, line 9: Also removed this as a separate title to directly have the full TOC expanded.

joshmoore (approved):
https://zarr-specs--180.org.readthedocs.build/en/180/extensions/data-types/complex/v1.0.html

Screen Shot 2022-11-25 at 09 20 41


That's _exceedingly_ clear. Open to opinions on if it makes reading difficult.

jstriebel:
@joshmoore I saw that #168 also accidentally updated the theme. There were some nice additions, but some things also felt off (e.g. the content was too narrow on large screens IMO). You can see version after the update [https://zarr-specs.readthedocs.io/en/latest/core/v3.0.html], and the version with the changes from this branch [here](https://zarr-specs--180.org.readthedocs.build/en/180/core/v3.0.html). If the updated theme does not fit, we could also pin the previous version, which was `0.8.0`.

I also used a more light-weight draft watermark now:
Screenshot from 2022-11-25 15-23-19

What do you think about the changes @joshmoore @MSanKeys963?

MSanKeys963 (approved):

MSanKeys963:
Hi @jstriebel. Thanks for sending the PR.

I had a look at the current version and one in this PR. I prefer the one in this PR [here](https://zarr-specs--180.org.readthedocs.build/en/180/core/v3.0.html). This one is less narrow and better utilises the spaces on the left and right-hand side of the screen. This may be a personal preference, and I'm happy to see what others think.

joshmoore (approved):

Issue #154, opened 2022-08-10
Beyond consolidated metadata for V3: inspiration from Apache Iceberg


rabernat:
I just reviewed the V3 spec. As with V2, the core spec does not address how to deal with unlistable stores. In V2, we developed the semi-ad-hoc "consolidated metadata" approach. I know there is already an issue about this vor V3 (#136), but I am opening a new one with the hope of broadening the discussion of what consolidated metadata can be.

## Background: What is Apache Iceberg

I have recently been reading about [Apache Iceberg](https://iceberg.apache.org/).  Iceberg is aimed at tabular data, so it is not a direct alternative or competitor to Zarr. Iceberg is not a file format itself (it can use Parquet, Avro, or ORC for individual data files). Instead, it is a scheme for organizing many data files individual into a larger (potentially massive) single "virtual" table. In addition to the table files, it tracks copious metadata about the files. In this sense, it is similar to Zarr: Zarr organizes many individual chunk files into a single large array. 

For folks who want quickly come up to speed on Iceberg, I recommend reading the following documents in order:
- [What is Iceberg? Features and Benefits](https://www.dremio.com/resources/guides/apache-iceberg/) (via Dremio)
- [Iceberg Spec](https://iceberg.apache.org/spec)

Architecturally, these are the main points

> This table format tracks individual data files in a table instead of directories. This allows writers to create data files in-place and only adds files to the table in an explicit commit.
> 
> Table state is maintained in metadata files. All changes to table state create a new metadata file and replace the old metadata with an atomic swap. The table metadata file tracks the table schema, partitioning config, custom properties, and snapshots of the table contents. A snapshot represents the state of a table at some time and is used to access the complete set of data files in the table.
> 
> Data files in snapshots are tracked by one or more manifest files that contain a row for each data file in the table, the file’s partition data, and its metrics. The data in a snapshot is the union of all files in its manifests. Manifest files are reused across snapshots to avoid rewriting metadata that is slow-changing. Manifests can track data files with any subset of a table and are not associated with partitions.
> 
> The manifests that make up a snapshot are stored in a manifest list file. Each manifest list stores metadata about manifests, including partition stats and data file counts. These stats are used to avoid reading manifests that are not required for an operation.

This diagram says 1000 words.

<img src="https://docs.dremio.com/images/iceberg-architecture.png" width="70%" />

Another interesting part of the spec, regarding the underling storage system capabilities:

> Iceberg only requires that file systems support the following operations:
> 
[...]

PR #182, opened 2022-11-25
Global storage transformers

rabernat:
As discussed in the ZEP call last week, this is a first pass at creating a "global storage transformers" concept, to be applied at the store (rather than array) level.

There are a few things that need cleaning up here:
- [ ] Double check that all RST references point to the right place
- [ ] Add more detail and example use-cases of both global and array storage transformers
- [ ] Add an example of a global storage transformer in `zarr.json`

But before doing that, I thought I would solicit some feedback on this. Does anyone see any fundamental problems with this concept? My motivating use case is #154. 

My main open question is: **when both array and global storage transformers are defined, what determines the order in which they are applied?** The only feasible choices appear to be to apply all the global transformers either _before_ or _after_ the array transformers. Should we make this an option in the protocol?

### DIFF
--- a/docs/core/v3.0.rst
+++ b/docs/core/v3.0.rst
@@ -445,12 +445,14 @@ terminology for a use case of reading from an array:
 
 *Storage transformer*
 
-    To provide performance enhancements or other optimizations,
+    To provide performance enhancements, optimizations, or additional features,
     storage transformers may intercept and alter the storage keys and bytes
-    of an array_ before they reach the underlying physical storage.
+    of a store before making queries against the physical storage layer.
     Upon retrieval, the original keys and bytes are restored within the
-    transformer. Any number of `predefined storage transformers`_ can be
-    registered and stacked. In contrast to codecs, strorage transformers can
+    transformer.
+    Storage transformers can act globally, for an entire store, or for an
+    individual array_. Any number of `predefined storage transformers`_ can be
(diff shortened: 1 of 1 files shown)


docs/core/v3.0.rst
```diff
@@ -1687,21 +1705,17 @@ Let "+" be the string concatenation operator.
...
-The stored transformed data is restored to its original state whenever data is requested
-by the Array. Storage transformers can be configured per array via the
-`storage_transformers`_ name in the `array metadata`_. Storage transformers which do
-not change the storage layout (e.g. for caching) may be specified at runtime without
-adding them to the array metadata.
+The purpose of storage transformers is to insert custom logic between the physical storage
+layer and the high-level Zarr data model.
+A storage transformer implements the same `abstract store interface`_ as a store_ and thus
+appears functionally identical to a store from the perspetive of the higher layers of the stack.
+However, it does not permantently store any information necessary to retrieve the original data.
```
shoyer, line 1712:
```suggestion
However, it does not permanently store any information necessary to retrieve the original data.
```

jstriebel:
Depends on #192

joshmoore:
Now conflicting.

jstriebel:
@rabernat Could you adapt this PR to cover group storage transformers for the current spec?

rabernat:
On today's call, we decided that we would defer this to an extension after the core v3 spec has been finalized.

Issue #176, opened 2022-11-24
v3 core spec: purl urls as identifiers


jstriebel:
Citing from @jbms's and @manzt's comments in https://github.com/zarr-developers/zarr-specs/pull/149#discussion_r929162112:

> Currently zarr v3 uses a mixture of full URLs and short strings as identifiers. I can see that there are some advantages to using full URLs: rely on existing DNS registration mechanism so there is no need for a separate "name reservation" process, automatic linking to a corresponding explanatory document assuming that the URLs actually remain valid.
>
> However, I think the long URLs would be cumbersome for a user to actually type manually, and even just looking at the metadata, it will be harder to read at a glance, since e.g. all of the codecs will start with the same long prefix, and just the part at the end is relevant. There are also surely a lot of cases where the user will want to specify something that is currently identified only by a URL, e.g. the codec to use, the metadata format to use, the grid to use, the storage transformer to use, etc.
>
> In order to save users from having to type out full URLs to identify these things, implementations may be tempted to invent their own short identifiers as well, and it may be confusing if these identifiers conflict, and there will be the added complexity of having to relate the short identifiers to the full URLs.

> It depends on the implementation, but my guess is that users often won't type these full URLs directly (moreover the implementation will likely fill in this metadata).

> It would be nice not to have to define an alternative more concise representation of the zarr v3 metadata, as that would likely lead to added confusion.
>
> More generally, users will need some way to specify everything that is in the metadata when creating a zarr repo/array, ideally in a concise way. If the zarr core spec or extension specs do not provide a concise representation, then each implementation may need to invent its own concise representation, which would be unfortunate.

The following questions seem to arise:
* Do we want to keep the purl URLs as identifiers in the metadata?
* If yes, should there be a common shorthand format for these?
* Technical question: Who updates and maintains these URLs? cc @joshmoore @MSanKeys963 @alimanfoo

My proposal is to keep the purl URLs, and recommend a shorthand format for implementations (the actual metadata must have the full URL, but implementations might use the common shorthands in their API), e.g. `https://purl.org/zarr/spec/codec/gzip/1.0` <-> `codec/gzip/1.0`, dropping the `https://purl.org/zarr/` prefix for URLs that are hosted under this prefix.

jstriebel:
[...]

PR #183, opened 2022-11-28
Change zarr_format to be just "zarr.dev/v3"

jbms:
Previously, the specification stated that `zarr_format` should be "https://purl.org/zarr/spec/core/3.0", with the last component indicating the major and minor version number, but with the requirement of forward and backward compatibility across different minor versions.

However, given the compatibility requirement, it is not clear what purpose the minor version number serves.

Furthermore, due to the fact that the zarr repository may be created with one zarr implementation, and later an array or group may be added using a different zarr implementation/version, it is not clear in which cases the version number may be updated.

Therefore, this commit removes the minor version, and changes to a more concise URL in order to make the metadata more easily read and written by humans.

### DIFF
--- a/docs/core/v3.0.rst
+++ b/docs/core/v3.0.rst
@@ -915,19 +915,7 @@ containing the following names:
     A string containing the URI of the Zarr core
     specification that defines the metadata format. For Zarr
     hierarchies conforming to this specification, the value must be
-    the string "https://purl.org/zarr/spec/core/3.0".
-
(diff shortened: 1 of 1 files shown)


jstriebel:
Related issue is #176.

I'm in favor of dropping the minor version number (but also see no harm in keeping it). I'd prefer to have consistent ids throughout the metadata, e.g. also in codecs etc…

jstriebel:
Superseded by #204, closing for now.

Issue #169, opened 2022-11-18
Issue overview: finalize and verify extension, store and codec mechanisms


jstriebel:
This is a high-level issue to track the progress around extensions, stores and codecs to finalize the v3 core spec. Finding the right abstraction levels for those is important, since it provides hints for implementors where to generalize concepts, which might make it easier in the long run to add future extension. In short: Having good extension points helps to grow support for extensions. Personally, I think the current status of the extensions is great, this is just to finalize those.

Please use or open the specific issues to discuss single points here. This is mostly to meta-discuss the overall strategy and overview, please feel free to recommend items that should be on this list, this just represents my current thoughts and neither exhaustive nor all strictly necessary (although I'll try to update it based on progress and comments):

### Ideas for other extension points
Those are ideas that were floating around in meetings and issues, we might want to add those extension points
* *WIP* Global storage transformer extension mechanism. PR #182
* [x] Metadata conventions could help to standardize metadata fields. Metadata mostly comes and is used by
  1. the writing zarr-lib itself
  2. higher-level libraries and viewers (eg. GeoZarr, OME-NGFF, neuroglancer, webKnossos, …)
  3. the user.
  
  Having a specific section and/or document for metadata convention would foster cross-compatibility between producers and viewers (i. & ii.). I think this should not enforce anything on the zarr level, but is a guideline/convention which might be used implementations. This also popped up in #139 and #144. I'll try to draft a PR for this to have some basis for discussion.

  See [ZEP 4](https://github.com/zarr-developers/zeps/pull/28)

### Finalize a spec per extension/store/codec:
To verify that all extension points are specified in full depth we might want to have finalized examples for each of those, also for stores and codecs. It might be useful to either include some of them directly into ZEP 1, or directly open separate ZEPs.
* [ ] ~Finalize data type extension~ (should be done in a different ZEP)
* [x] Finalize one store spec,
  probably the [file system store](https://zarr-specs.readthedocs.io/en/latest/stores/filesystem/v1.0.html)
  see #195.
* [x] Finalize at least one codec, maybe simply all current ones at https://zarr-specs.readthedocs.io/en/latest/codecs.html
  See #87, #187
* [ ] ~Add v2 compatibility extension~
* [x] Clarify dependencies between extension points. E.g. must per-array-extensions be mentioned in the entrypoint metadata? Must data-types, storage transformers etc. be listed in the `extensions` of the array?
  See #191

### Add templates
[...]

PR #187, opened 2022-11-30
move codecs into separate (versioned documents), update urls

### LINKED ISSUE #87: v3 Spec: Finalize current codecs
One of the TODO items in spec v3 is the process around adding codec.
Once https://github.com/zarr-developers/zarr-specs/pull/12 is resolved a link should be added in V3 "Chunk encoding" section.

### LINKED ISSUE #148: Versioning codecs
Currently codecs don't have a version. Thus if they are changed in a breaking API, this would break loading the data. Would it make sense to include one? Or would we want to handle this a different way (like naming the codec differently, for example `bz2` & `bz2_2`)?

xref: https://github.com/fsspec/kerchunk/pull/198#issuecomment-1191726825 (where this came up originally)

cc @martindurant @joshmoore

jstriebel:
This PR
* separates the codec specifications into separate documents similar to extensions or stores
* uses versions for the codecs, similar to all other specifications (fixes #148)
* added all codecs to ZEP 1(part of #169, fixes #87)
  related ZEP PR: https://github.com/zarr-developers/zeps/pull/23
* updates purl urls to reflect the structure of the specification documents (but might change more anyways later, see #176)
* updates github links to point to actual documents
* fixes some references
* removed the `filters` extension which is superseded by the `codecs` list as of #153

**Note**: The text of the actual codec specifications is unchanged.


jbms:
array-extensions/filters can be removed since that functionality is now covered by the `codecs` metadata field

jbms:
As also mentioned in #183, I think there is value in keeping the metadata representation concise, because in some cases users may read and write it directly.  For example, some zarr implementations (like tensorstore) may just rely on the metadata json representation itself for specifying codecs, rather than defining a separate programming API for each codec.  I think we might look at HTML as a model: we have `` not `<https://purl.org/html/tags/img/v7.3>`  As new features are added, browser compatibility is documented:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#browser_compatibility
https://caniuse.com/mdn-html_elements_img_crossorigin

My own preference is that, for example, the gzip codec is specified with just plain "gzip".  If we later need to make an incompatible change (unlikely in my opinion), we could then introduce a "gzip2" or "gzip/v2" codec as well.  If a backwards-compatible change is added, such as introduction of a new optional field, then we could continue using just "gzip", and implementations that don't support the new field would either fail or ignore it, depending on how the codec is defined.  I can see that there is one downside --- with an explicit version number, if you try to read "gzip/v1.3" and the implementation is only aware of "gzip/v1.2", then the error message can say "gzip/v1.3" not supported, latest version supported is "gzip/v1.2", as opposed to "Invalid field 'foobar', array may have been created with a newer version of the gzip codec.".

Only codecs defined by external organizations, and that haven't been standardized as part of the zarr specification, would use a longer identifier (e.g. a full URL, possibly without the scheme).

It still could make sense to have version numbers for the codec specification itself, so that implementations can specify in their documentation exactly which version of the codec specification they claim to support.  The proposed mechanism here seems to be to duplicate the file under a new name with an incremented version number when any change is made.  I wonder if we could instead just rely on git to track multiple versions; implementations could then just link to the older spec file on github.  There are also various ways to manage multiple versions in the generated documentation.

jstriebel:
@jbms I agree with you in all points, but I still would prefer to split up the codec documents. I just reconciled the purl URLs and would like to change or update them, see #176. How about continuing the URL/id conversation there? I think it should not block this PR, this just cleans up lot's of things to have a clean overview now.


> array-extensions/filters can be removed since that functionality is now covered by the `codecs` metadata field

Good point, removing it in this PR as well now.

@jbms Would you like to see any further changes except for an ongoing discussion about the purl urls? As explained, I'd rather defer that discussion to #176 for now.

jbms (approved):

PR #189, opened 2022-11-30
Replace chunk_memory_layout with transpose codec

jbms:


docs/core/v3.0.rst
```diff
@@ -757,16 +697,8 @@ To encode or decode a chunk, the encoded and decoded representations for each
...
-   follows:
-
-   - If ``chunk_memory_layout`` is equal to ``"C"``, the shape is equal to the
-     chunk shape.
-   - If ``chunk_memory_layout`` is equal to ``"F"``, the shape is equal to the
-     chunk shape, with the dimension order reversed.
-   - If ``chunk_memory_layout`` is defined by an extension, the extension
-     defines the shape.
+   multi-dimensional array with the same data type as the zarr array, and shape
+   equal to the chunk shape.
```
jstriebel, line 701: How about adding this?
```suggestion
   equal to the chunk shape. This is usually referred to as row-major or C order layout. To change
   the underlying memory layout without changing the order in the metadata, please use the
   :ref:`transpose-codec-v1`.
```

jstriebel: (didn't test if the reference works like this)

jstriebel (commented):
Thanks a lot @jbms, this mostly LGTM :+1:
Please also add short notes to the `Comparison with Zarr v2` and `Change log` sections.

jstriebel (dismissed):
LGTM, thanks :+1:

jbms:
@jstriebel This needs your review again now that I've resolved the conflict.

jstriebel (approved):

rabernat:
Sorry I never provided a review on this. FWIW, I finally looked it over and I think it's 👌 . Thanks @jbms and @jstriebel for your work here.

Issue #177, opened 2022-11-24
v3 core spec: Consider to drop /meta prefix, have file at URI


jstriebel:
Citing @jbms from https://github.com/zarr-developers/zarr-specs/pull/149#discussion_r929356279:

> As discussed on the community meeting, this naming scheme has the drawback in that there is not a good way to have a path directly to a non-root array.  Additionally, it was noted that to better integrate with existing filesystem completion in editors, etc., it would be helpful if the path were a real filesystem path.
>
> A few proposals were made:
> 
> - Instead of `"meta/root" + P + ".group.json"` as the key for the metadata file, instead have it just be: `P + ".zr3"`.  This would lead to a key of just `".zr3"` for the root.  It was noted though that with a consolidated metadata extension these metadata files would not actually exist.  The data could be placed in e.g. "_data" or something, where names beginning with an underscore could be reserved to prevent conflicts.  This would still require a way to locate the root directory --- that could be done by storing the relative path inside the array metadata.
> - Alternatively, we could leave the directory structure alone but require an extension as part of the directory containing the root, e.g. "foo.zr3", and disallow any array or group names from ending in ".zr3".  Then you could use: "path/to/root.zr3/path/to/array" as a pseudo-path to an individual array.  The downside is that it may be confusing to use something that looks like a path, and where a portion corresponds to a real filesystem path but a portion does not.  File completion in editors also wouldn't support this.
> - We could use a special syntax to combine both a path to a root and a path to an array into a single string, e.g. "path/to/root//path/to/array" or "path/to/root#/path/to/array".  We would need to carefully choose the syntax to avoid conflicts with e.g. fsspec, and file completion in editors also wouldn't support this.

Some more comments from discussion rounds I remember:
* .json suffix is useful to have a correct mimetype by default for many stores, e.g. S3
* Having a URI (TBD, see #132) to an array or group, it would be great to actually find a directory or file there. E.g. `s3://bucket-name/key-name/name-of-the-zarr-path.zarr/hierarchy/path/my-data.array.json` could be a URI to point to the `my-data` array at the path `hierarchy/path/my-data` of the zarr hierarchy which is placed under `s3://bucket-name/key-name/name-of-the-zarr-path.zarr/`. (Just made up a URI here as an example, feel free to discuss this in #132). Using such a URI schema and dropping the `/meta` prefix, one could find the relevant file (at least for filesystem or http stores or using appropriate clients for other stores).
[...]

PR #196, opened 2022-12-22
node names: allow unicode strings, recommend subset

### LINKED ISSUE #56: Node name character set

### LINKED ISSUE #114: Name conventions for zarr
I gather from the V2 spec that the notion of a legal name for arrays, groups, and attributes is taken wholesale from h5py, which in turn means that the HDF5 name rules are used.
Is this correct?

jstriebel:
Applies the latest suggestions from issue #56, allowing more flexible node names and just recommending a safe subset. Feel free to add or propose changes, this is just to get things startet.
Fixes #56, fixes #114.


docs/core/v3.0.rst
```diff
@@ -455,29 +455,22 @@ Node names
...
-which is a string of characters. To ensure consistent behaviour
-across different storage systems, the following constraints apply to
+which is a string of characters. The following constraints apply to
 node names:
 
-* must not be the empty string ("")
+* must be a Unicode code point sequence
+* must not be the empty string (``""``)
+* must not include the character ``"/"``
+* must not start with the reserved prefix (TBD, e.g. ``"_z_"``)
```
jstriebel, line 464: TBD, see also comments here: https://github.com/zarr-developers/zarr-specs/issues/177#issuecomment-1362891380

jstriebel:
@jbms I updated this PR with the latest discussion points from #56 and #177, do you think it covers the current points?

ethanrd:
@jstriebel - There's still the issue of limiting the set of characters allowed.

Also, I think a Unicode encoding needs to be specified. UTF-8 seems a good choice as a default. (Could an extension(s) be added later to allow for other encodings?)

jbms:
> @jstriebel - There's still the issue of limiting the set of characters allowed.
> 
> Also, I think a Unicode encoding needs to be specified. UTF-8 seems a good choice as a default. (Could an extension(s) be added later to allow for other encodings?)

Yes, currently the spec is not clear on whether the keys of a store are logically considered byte strings or unicode strings.  With the previous restrictions on keys that distinction did not matter, but with those restrictions removed it is important.  If they are logically considered byte strings, then we would need to specify that Unicode is encoded as UTF-8, and there is also the possibility of storing strings that aren't valid Unicode.

If the keys are logically considered Unicode strings, then encoding is not really a concern of the core spec, and may only be relevant in defining a specific store (e.g. filesystem store).

In general I would say that our choice should be informed by how the relevant store implementations actually work:

- Linux filesystem: byte strings, but usually UTF-8 encoded these days
- macOS filesystem: Unicode strings
- Windows filesystem: Strings of 16-bit numbers, not necessarily valid UTF-16.  However, we may wish to ignore that possibility and just say it is Unicode strings.
- S3: Unicode strings
- GCS: Unicode strings
- Azure blob storage: Unicode strings
- ZIP file: Unicode strings (approximately)
- LMDB: Byte strings
- Git: Unicode strings

This is isn't an exhaustive list of possible store implementations, but I think provides sufficient evidence in favor of defining store keys as unicode strings rather than byte strings.

jstriebel:
Thanks for the @ethanrd reminder and @jbms for the research! I added a note about the UTF-8 encoding as a default, which seems appropriate to include but doesn't overspecify things if stores/implementations natively use Unicode strings. I also disallowed the Cf/Cc characters, forgot about that before.

docs/core/v3.0.rst
```diff
@@ -441,38 +441,57 @@ Node names
 
 The root node does not have a name and is the empty string ``""``.
 Except for the root node, each node in a hierarchy must have a name,
-which is a string of characters. To ensure consistent behaviour
-across different storage systems, the following constraints apply to
+which is a string of characters. The following constraints apply to
```
jbms, line 444: I would say "Unicode code points" rather than "characters" here, and then you don't need to specify "unicode code point sequence" below.

jbms:
I am a bit hesitant to attempt to put restrictions on the Unicode characters at this point, such as excluding Cc and Cf character categories, because it adds some implementation complexity to validate that and it is not clear what problem we are trying to solve with that.  If the goal is to try to avoid names that will be ambiguous or confusing when printed to a terminal, then I think a much larger set of exclusions would be needed, and in general it seems like restrictions such as that might better be added as part of the normalization extension.

Additionally, one issue that comes up with excluding Unicode character classes is that later versions of the Unicode spec may introduce new characters within those classes.  Therefore, to ensure that this rule doesn't become violated with a new Unicode version, you must also exclude any characters that are unassigned as of the latest Unicode version known to the implementation.  This is something that the Apple filesystem APFS enforces, but also seems like something that would be more appropriate within the normalization extension.

ethanrd:
I agree an extension is a good place for all the Unicode details. I think the goal for these Unicode character recommendations are the same as for the `[a-zA-Z0-1-_.]` recommendation, consistent behaviour between implementations. So perhaps just shorten the "When using non-ASCII Unicode characters" sentence and leave the details for the extension. Something like:

> When using non-ASCII Unicode characters, we similarly recommend the use of a limited set of characters that are detailed in the Unicode extension.

ethanrd:
Regarding Unicode adding characters in the future, the "Default Identifiers" section of the "Unicode Identifier and Pattern Syntax" ([[https://www.unicode.org/reports/tr31/][UAX 31]]) document says (emphasis mine):

> "The formal syntax provided here captures the general intent that an identifier consists of a string of characters beginning with a letter or an ideograph, and followed by any number of letters, ideographs, digits, or underscores. It provides a definition of identifiers that is **guaranteed to be backward compatible with each successive release of Unicode**, but also adds any appropriate new Unicode characters."

jstriebel:
Thanks @jbms & @ethanrd for the input. I agree that disallowing format & control categories goes too far in the core spec. Using a recommendation that reflects more than just normalization also is needed. Instead of recommending "Default Identifiers", I went forward and added a recommendation for "Immutable Identifiers". Those ensure that any future unicode strings which were valid stay valid, which is not the case for Immutable Identifieres. From [UAX31 – Immutable Identifiers](https://www.unicode.org/reports/tr31/#Immutable_Identifier_Syntax):

> The disadvantage of working with the lexical classes defined previously is the storage space needed for the detailed definitions, plus the fact that with each new version of the Unicode Standard new characters are added, which an existing parser would not be able to recognize. In other words, the recommendations based on that table are not upwardly compatible.

Having this as a recommendation seems appropriate. This still seems quite strict, but probably ok for now ([pattern_syntax](https://www.sql-und-xml.de/unicode-database/pattern_syntax.html) is quite a lot):
> Define identifiers to be any non-empty string of characters that contains no character having any of the following property values:
> * Pattern_White_Space=True
> * Pattern_Syntax=True
> * General_Category=Private_Use, Surrogate, or Control
> * Noncharacter_Code_Point=True


PS: Without specifying an additional profile the "Default Identifiers" seem too prohibitive in our use case, e.g. strings may not start with `_` then. 

What do you think @jbms @ethanrd?

ethanrd:
This bit of [UAX31 - Immutable Identifiers](https://www.unicode.org/reports/tr31/#Immutable_Identifier_Syntax) gives me pause:
> The drawback of this method is that it allows “nonsense” to be part of identifiers because the concerns of lexical classification and of human intelligibility are separated. Human intelligibility can, however, be addressed by other means, such as usage guidelines that encourage a restriction to meaningful terms for identifiers. For an example of such guidelines, see the XML specification by the W3C, Version 1.0 5th Edition or later [XML].

Since this is a recommendation rather than a requirement, I wonder if avoiding "nonsense" isn't more important than backward/upward compatibility.

Either way, I think both the Immutable and the Default Identifier syntax exclude the period ('.') and there may be other characters that should be added to the allow list.

Also, I realize I've been thinking of this more as a recommendation to users more than implementors, similar (I think) to the `[a-zA-Z0-9-_.]` recommendation. Implementations and stores will support Unicode to varying degrees but to be most interoperable across those implementations and stores it is recommended that users limit the characters in the names they give to their groups and variables and such. Implementations might ignore this recommendation or they might give a warning or error if a user tries to name a node with a string that doesn't match the recommendation.

Normalization, on the other hand, is advise to implementations. Basically, don't try to compare names until those names have been normalized using the same normalization scheme.

Does the Zarr specification differentiate between user and implementor recommendations/guidance?

jstriebel:
> Either way, I think both the Immutable and the Default Identifier syntax exclude the period ('.') and there may be other characters that should be added to the allow list.

You're right, and I think I've finally found an appropriate recommendation:
[UTS39: section General Security Profile for Identifiers](http://www.unicode.org/reports/tr39/#General_Security_Profile)


This is also recommended [Unicode Security Considerations – General Recommendations](https://unicode.org/reports/tr36/#Recommendations_General):
> When defining identifiers in programming languages, protocols, and other environments:
> 1. Use the general security profile for identifiers from Section 3, [Identifier Characters](http://www.unicode.org/reports/tr39/#Identifier_Characters) in UTS # 39: Unicode Security Mechanisms [[UTS39](https://unicode.org/reports/tr36/#UTS39)].
>    * Note that the general security profile allows characters from [Table 3, Candidate Characters for Inclusion in Identifiers](http://www.unicode.org/reports/tr31/#Table_Candidate_Characters_for_Inclusion_in_Identifiers) in [[UAX31](https://unicode.org/reports/tr36/#UAX31)], such as U+00B7 (·) MIDDLE DOT used in Catalan.
> 2. For equivalence of identifiers, preprocess both strings by applying NFKC and case folding. Display all such identifiers to users in their processed form. (There may be two displays: one in the original and one in the processed form.) An example of this methodology is Nameprep [[RFC3491](https://unicode.org/reports/tr36/#RFC3491)]. Although Nameprep is currently limited to Unicode 3.2, the same methodology can be applied by implementations that need to support more up-to-date versions of Unicode.

Point 2 recommends NFKC again, which would differ from NFC of filesystems (see https://github.com/zarr-developers/zarr-specs/issues/56#issuecomment-1387593918). However, the general security profile disallows non-NFKC normalized characters. I tend to simply recommend case-folding and NFKC normalization here, which would exactly follow all recommendations. The technical comparison is still case-sensitive.

This seems to be a better fit than UAX#31, which seems to be targeted towards identifiers in programming languages.
The newest commit reflects those changes.

 I would very much like to get this merged soon so that the [implementation councils can vote on it for provisional acceptance](https://github.com/zarr-developers/zeps/discussions/24). Please feel free to add suggestions directly in the spec changes of this PR.

> Does the Zarr specification differentiate between user and implementor recommendations/guidance?

The specification is directed to implementors mostly. It only contains hints what implementations might recommend to end users (via docs, warnings, etc). I assume few users will read the specification. In this case I uses `we recommend users to …`, and also add this to the `[a-zA-Z0-9-_.]` recommendation.

jbms (approved):

jstriebel:
@ethanrd I'll go forward here and merge this for now, but feel free to still add feedback, happy to add changes in a separate PR if necessary.

Issue #192, opened 2022-12-07
Consider to remove entrypoint metadata, do not specify root


jstriebel:
During the last ZEP meeting we discussed if it's possible and useful to not have an explicit root of a zarr hierarchy. See the [meeting notes here](https://zarr.dev/zeps/meetings/2022-12-01.html).

This would include to drop the global `/meta` and `/data` prefix as described in #177, but possibly also the entrypoint metadata.

This discussion is directly related to those comments: https://github.com/zarr-developers/zarr-specs/issues/177#issuecomment-1336690423

I'll try to summarize the arguments:

#### Pro explicit root
* A zarr hierarchy has a clear address.
* Global extensions which move/change metadata (e.g. consolidated metadata) can be configured via the entrypoint metadata. Entrypoint metadata can be a good "gateway" to set up arbitrary extensions, without needing to find any other metadata first.

#### Contra explicit root
* Groups and arrays can be moved anywhere and still are valid groups/arrays (but extensions would need to be part of all respective metadata). (Let's refer to this as **nesting invariance**.)
* Any extensions which don't affect the metadata can be (and then maybe should be) copied into all contained arrays.

-----------------------

One particular case that changes some of these arguments is introducing group-wise extensions, e.g. consolidated metadata within an arbitrary sub-group of a hierarchy. This would weaken nesting invariance, as only the whole group with the extension can be moved freely. E.g. moving any array in the group with consolidated metadata would require an update of the metadata. However, if extensions don't affect metadata, it would be very pratical to simply copy all extensions into the respective arrays.

---------------------

My currently favoured strategy would be:
* Remove the explicit entrypoint metadata
* Move the relevant metadata into the groups:
  * Probably drop metadata encoding (#174)
  * `zarr_format` in all group and array metadata
  * `extensions` in (relevant or all) group and array metadata
* Implicit groups are always opened via their next explicit parent group, arrays w/o metadata must look up extensions in their parent group
  * The topmost entry of a group hierarchy must be explicit (similar to current entrypoint), e.g. if a collection of arrays should be openable as a group, the topmost group must have metadate, e.g. to specify the zarr version.
  * The strategy for implementations would be to traverse the store hierarchy upwards from any given url, until array or group metadata is found.

-------------

cc all participants of the last zep call: @MSanKeys963 @joshmoore @jbms @WardF @DennisHeimbigner @rabernat

jstriebel:
[...]

PR #200, opened 2023-02-02
Restructure keys and remove entry-point metadata

### LINKED ISSUE #177: v3 core spec: Consider to drop /meta prefix, have file at URI

jbms:
Fixes https://github.com/zarr-developers/zarr-specs/issues/192, fixes #177

### DIFF
--- a/docs/core/v3.0.rst
+++ b/docs/core/v3.0.rst
@@ -225,9 +225,10 @@ The following figure illustrates the first part of the terminology:
 *Hierarchy*
 
     A Zarr hierarchy is a tree structure, where each node in the tree
-    is either a group_ or an array_. Group nodes may have children but
+    is either a group_ or an array_.  Group nodes may have children but
     array nodes may not. All nodes in a hierarchy have a name_ and a
-    path_.
+    path_.  The root of a Zarr hierarchy may be either a group_ or an array_.
+    In the latter case, the hierarchy consists of just the single array.
 
 .. _group:
 .. _groups:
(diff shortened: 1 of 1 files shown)


rabernat (dismissed):
I have reviewed this and think it looks all good.

Thanks for taking the time to write this down Jeremey.

jstriebel (dismissed):
Thanks @jbms! Also LGTM, just pushed a commit with minor formatting edits. I also started to clean-up the Storage – Operations section, but noticed that it's quite out of date and probably better fixed in a separate PR. (We might even consider removing it completly, I'm not quite sure how helpful it is.)

jstriebel:
@rabernat Could you approve this once again? my approval doesn't count since I commited and also my commit automatically discarded your previous one.

jstriebel:
PS: A changelog entry is missing, could you add one @jbms?

joshmoore (approved):
:+1:

PR #202, opened 2023-02-08
Add int64 and uint64 to core dtypes

jstriebel:
I just noticed that #155 dropped int64 and uint64 dtypes, which are re-added here.


jbms (approved):
Thanks

PR #203, opened 2023-02-09
Allow native storage formats for stores.

### LINKED ISSUE #90: Native metadata storage
Currently the v3 core protocol spec assumes that array and group metadata documents will be encoded in some way (default JSON) prior to storage. This allows use of stores like file systems or cloud object stores, because they don't need to be able to do anything other than store and retrieve a sequence of bytes.

However, some stores might be able to store metadata natively, i.e., without encoding. For example, a MongoDB store would be able to store the metadata documents natively. Similarly, an SQLite store might choose to store metadata natively, making use of the [JSON1](https://www.sqlite.org/json1.html) extension. 

There are several potential advantages to using these native storage capabilities for metadata. For example, both MongoDB and SQLite could manage concurrent modification to the user attributes on a zarr array or group. These storage engines could also support efficient queries across all metadata in a zarr hierarchy.

How should we deal with this in the v3 core protocol spec? I.e., how do we accommodate stores with native support for metadata storage, as well as stores that require metadata to be encoded?

jstriebel:
Proposal to fix #90. We might also choose to restrict non-byte-string formats to metadata documents only, or more specifically, to non-chunk-data only.

### DIFF
--- a/docs/core/v3.0.rst
+++ b/docs/core/v3.0.rst
@@ -1268,7 +1268,10 @@ The store interface defines a set of operations involving `keys` and
 `values`. In the context of this interface, a `key` is any string
 containing only characters in the ranges ``a-z``, ``A-Z``, ``0-9``, or
 in the set ``/.-_``, where the final character is **not** a ``/``
-character. A `value` is a sequence of bytes.
+character. In general, a `value` is a sequence of bytes. Specific stores
+may choose more specific storage formats, which must be stated in the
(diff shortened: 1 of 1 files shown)


joshmoore (approved):
Certainly no objection. As I mentioned in the discussion, the only real constraint I can think of currently is "it must be valid JSON after the conversion" which is perhaps a given.

jbms (approved):

PR #204, opened 2023-02-13
Restructure docs, clarify extension points and configs, use short ids

### LINKED ISSUE #176: v3 core spec: purl urls as identifiers

### LINKED ISSUE #191: Why do storage transformers need "type" separate from "configuration"
This is what V3 currently says about how to specify storage transformers

https://github.com/zarr-developers/zarr-specs/blob/b509f142e129ef2b13979044c9406755af2ff54e/docs/core/v3.0.rst?plain=1#L1165-L1174

Why can't we just put `type` _inside_ `configuration`? That just seems simpler. Plus, it may not make sense to define `type` for some storage transformers. That means `type` is a transformer-specific configuration parameter anyway.

cc @jstriebel

jstriebel:
This PR

* restructures the headers to integrate other zarr pages,
* adds links to v2 and v1 specs,
* clarifies extension points, fixes #191:
  * future features will be added as new top-level keys
  * extension points have coherent configs
  * updated the stability policy accordingly
* changes all identifiers to be simple short forms, fixes #176,
* adds minor cleanups, e.g. removing resolved questions.

Preview of the built docs:
https://zarr-specs--204.org.readthedocs.build/en/204/v3/core/v3.0.html


jbms:
Note: complex numbers are now part of the main spec so we can remove that extension.

docs/v3/core/v3.0.rst
```diff
@@ -1112,40 +1065,26 @@ A Zarr group metadata object must contain the following mandatory key:
...
-
-    If an implementation of this specification encounters an extension
-    that it does not recognize, but the value of ``must_understand``
-    is ``false``, then the extension may be ignored and processing may
-    continue. If the extension is not recognized and the value of
-    ``must_understand`` is ``true`` then processing must terminate and
-    an appropriate error raised.
-
-``descenents_may_have_group_extensions``
+``descenents_must_be_read``
```
jbms, line 1078: Note: My original name had a typo, "descenents" instead of "descendents".

However, I would actually propose that in the interest of getting v3 out sooner, we leave out this property for now so that we can consider it more carefully after getting out v3 but before adding any group extensions.  If we add this property later, and an implementation is unaware of it, it will fail if it encounters it, which is correct behavior, since if it doesn't know about this property then it also won't know about the group extensions.   There are some edge cases where technically the implementation might be able to read parts of the group successfully even without knowing about this property, but I don't think we have to worry about that.

docs/v3/core/v3.0.rst
```diff
@@ -1160,13 +1099,18 @@ Optional keys:
 For example, the JSON document below defines an explicit group::
 
     {
-        "zarr_format": "https://purl.org/zarr/spec/v3#group",
+        "zarr_format": 3,
```
jbms, line 1102: I think we need something to explicitly distinguish between arrays and groups, rather than having to do it based on the presence of data_type or other array properties.  Note that I had changed `zarr_format` to be different for arrays and groups ( "https://purl.org/zarr/spec/v3#array" vs "https://purl.org/zarr/spec/v3#group").

docs/v3/core/v3.0.rst
jstriebel: > However, I would actually propose that in the interest of getting v3 out sooner, we leave out this property for now so that we can consider it more carefully after getting out v3 but before adding any group extensions. If we add this property later, and an implementation is unaware of it, it will fail if it encounters it, which is correct behavior, since if it doesn't know about this property then it also won't know about the group extensions. There are some edge cases where technically the implementation might be able to read parts of the group successfully even without knowing about this property, but I don't think we have to worry about that.

Good idea! I added a separate issue for it: #207

docs/v3/core/v3.0.rst
jstriebel: I'd rather put that into a different key than the version, if there are no objections. How about `node_type`, which is either `group` or `array`?

jstriebel:
@jbms Thanks for the review! I applied your feedback, and also updated the sharding and codecs spec to the newest format. This is also necessary for the data-types, but I think those need to be partially re-written anyways, not only with the new configuration syntax. It would be great to have somebody authoring a ZEP for those. (cc @MSanKeys963, might be something to ask around for in the next community meeting?)

Also cc @joshmoore and @MSanKeys963, what do you think of the updated menu header? I could imagine to use the same headers with external links on multiple pages, to get the impression of a single coherent page. This doesn't need to be tackled in the specs first, but I thought this could be a good opportunity to play around with this idea, that's why I added it here.

docs/v3/core/v3.0.rst
```diff
@@ -844,18 +822,21 @@ encoded for storage.
...
+Each Zarr array in a hierarchy must have an array metadata document, named
+``zarr.json``. This document must contain a single object with the following
+mandatory names:
 
 ``zarr_format``
 ^^^^^^^^^^^^^^^
 
-    Must be the string "https://purl.org/zarr/spec/v3#array".
+    An integer defining the version of the storage specification to which the
+    array store adheres, muste be ``3`` here.
```
jbms, line 833: type on "muste"

docs/v3/core/v3.0.rst
```diff
@@ -844,18 +822,21 @@ encoded for storage.
...
 
-    Note that this value is given as a URI rather than as a simple
-    version number string to help with discovery of this
-    specification.
+``node_type``
+^^^^^^^^^^^^^^^
+
+    A string defining the type of hierarchy node element, muste be ``array``
```
jbms, line 838: typo on "muste"

docs/v3/core/v3.0.rst
```diff
@@ -1112,43 +1074,21 @@ A Zarr group metadata object must contain the following mandatory key:
...
 ^^^^^^^^^^^^^^^
 
-    Must be the string "https://purl.org/zarr/spec/v3#group".
+    An integer defining the version of the storage specification to which the
+    array store adheres, muste be ``3`` here.
+
+``node_type``
+^^^^^^^^^^^^^^^
+
+    A string defining the type of hierarchy node element, muste be ``group``
```
jbms, line 1083: typo on "muste"

jbms (dismissed):
Aside from the typos, looks good to me.  We may want to iterate more, but easier if this is submitted first, I think.

PR #210, opened 2023-02-20
updated terminology images

jstriebel:


jstriebel:
Thanks for the review @rabernat!

> * 2x smaller

 Fixed now :+1:

> * not on a white background

The old ones where on a transparent background, the new ones have a white one which is better for dark themes.

I also incorporated the changes from PR #214 to the example metadata in the hierarchy image.

rabernat (approved):

Issue #132, opened 2022-02-08
v3: Define standard "URL" syntax for referencing a specific array, group, attribute within a zarr repository


jbms:
It is convenient to be able to reference an array (or group or attribute) with a single string.

Existing zarr v2 implementations have invented various URL syntaxes.  E.g. in Neuroglancer:

`zarr://https://whatever/path/to/array`

(Arguably a better syntax would have been: `zarr+https://whatever/path/to/array`)

With zarr v3 it is necessary to separately specify the root and the path within the repository.  To allow for better interoperability, it would be helpful for zarr v3 to specify a URL syntax to avoid divergence between implementations.  Of course the zarr specification itself does not concern itself with the underlying storage mechanism such as https, but it would still be helpful to standardize a URL syntax.

rabernat:
Good idea! Pinging @martindurant for fsspec URI interoperability angle.

jbms:
A lot of different tools operate on arrays, and it is convenient to be able to specify the location of an array in a form that facilitates easy communication/interoperability between different people and tools.

For example, it is convenient to have a "specifier" (be it a "URL" or something else) that can be copy-and-pasted directly between email messages, chat messages, command-line arguments, UI text entry boxes in various tools, etc.

Certainly in principle this specifier does not have to have "URL-like" syntax, e.g. could be a JSON representation such as:

```
["https://host/path/to/zarr_repo", "/path/to/array"]
```

or

```
{"driver": "zarr3", "kvstore": "https://host/path/to/zarr_repo", "array": "/path/to/array"}
```

But I would say there are a lot of advantages to using URL-like syntax:

- Some existing software like Neuroglancer already uses URL syntax to identify data sources, and if there is no standardized URL syntax for zarr, one will be invented anyway.
- Users are already accustomed to URLs and will likely be more readily able to copy and paste them without accidentally munging them.
- A single string without quotes, spaces or other problematic characters is much easier to specify as a command-line argument.

In general the same reasons why URL syntax is convenient for web browsers and fsspec itself also apply to specifying a zarr array.

As for the "zarr://" part (i.e. specifying that the format is zarr), obviously that isn't necessary for software that only supports zarr, and perhaps they would exclude the "zarr+" or "zarr://" prefix.  But a lot of tools (e.g. Neuroglancer) support multiple formats, and therefore need a way to specify the format as part of the URL.

[...]

PR #211, opened 2023-02-22
Various storage updates

### LINKED ISSUE #137: recommended file extension for zarr-v3 stores (.zr3?)
Currently the spec does not specify how a v3 store should be named. I am curious whether we should either encourage or enforce use of a specific suffix like `.zr3` (currently used for the folder names for v3 cases in the `zarr_implementations` tests) or `.zarr3`.

If this was enforced, we could infer from a file path which version of the Zarr protocol the store was created with without having to provide a `zarr_version=3` kwarg (or implementing some kind of inspection of the file hierarchy to try and autodetect the version). 

Currently in the draft implementation of  https://github.com/zarr-developers/zarr-python/pull/898, no such autoinspection is attempted. If a function takes a store argument and a `StoreV3` subclass is passed in, it is safe to assume version 3. Otherwise version 2 is assumed for backwards compatibility unless a `zarr_version=3` kwarg is passed in.

### LINKED ISSUE #206: Update or remove "Storage – Operations" section
The [Storage – Operations](https://zarr-specs.readthedocs.io/en/latest/core/v3.0.html#operations) section is currently not up to date. We should either update it or remove it.

jstriebel:
* Updates the `Storage – Operations` section to reflect the current state of the spec (fixes #206)
* Adds a recommendation for a `.zarr` suffix (fixes #137)
* Adds a canonical URI format to stores (part of #132)
* Added a note about race-conditions for implicit stores (as noted by @jbms in some other issue I can't remember)
* Fixed mermaid rendering (the path to the js file at the CDN changed in the latest version, using self-hosting instead)


docs/v3/stores/filesystem/v1.0.rst
```diff
@@ -182,6 +182,19 @@ in the section above.
   directory path ``dp``.
 
 
+Canonical URI
+=============
+
+The canonical URI format for this store follows the file URI scheme defined in
+[RFC8089]_. In the common case where the hostname is omitted this is
+``"file:///" + P``, where `P` is the base directory path, e.g.
+``"file:///C:\\data\\foo\\bar"``.
```
jbms, line 191: This should be `"file://" + P`, e.g. `file://c:\data\foo\bar` or `file:///home/user/path/to/data.zarr`

Also, the backslashes inside the double quoted url is a bit confusing here, because we haven't specified what escaping mechanism we are assuming for that double quoted string.

If we intend to later support more elaborate URL syntax, e.g. `file:///path/to/group.zarr#path/to/array`, then we also need to define how special characters like "#" are escaped.  Presumably we would want to use percent encoding, as is typical for URLs.  But means that the following comment:

> In cases where a filesystem store may be considered the default, the ``"file:///"`` prefix can be omitted and only the base directory path is used

is problematic, because for a bare path you would presumably not use percent encoding.

jstriebel: >  e.g. file://c:\data\foo\bar or file:///home/user/path/to/data.zarr

It seems like for windows it should even be `file:///c:/data/foo/bar`.

> Also, the backslashes inside the double quoted url is a bit confusing here, because we haven't specified what escaping mechanism we are assuming for that double quoted string.

I was assuming that those need to be escaped for RST, but this is not the case for code formatted parts.

> If we intend to later support more elaborate URL syntax, e.g. file:///path/to/group.zarr#path/to/array, then we also need to define how special characters like "#" are escaped. Presumably we would want to use percent encoding.

Yes, this is even specified by the file URI RFC, as it relies on the default URI RFC.

I'll adapt those points and have a rather flexible definition of the form without the prefix.

docs/v3/core/v3.0.rst
```diff
@@ -1129,6 +1129,9 @@ interface`_ subsection. The store interface can be implemented using a
 variety of underlying storage technologies, described in the
 subsection on `Store implementations`_.
 
+Additionally, a store should specify a canonical URI format that can be used to
+identify nodes in this store. Implementations should use the specified formats
+when opening a Zarr hierarchy to automatically determine the appropriate store.
```
rabernat, line 1134: I don't understand what this means. Could we give an example?

docs/v3/core/v3.0.rst
```diff
@@ -1396,125 +1399,111 @@ array, the prefix is the empty string.  For a non-root array with hierarchy path
...
+    the array metadata stored at ``meta_key(P)``.
 
 **Discover children of a group**
 
     To discover the children of a group at hierarchy path `P`, perform
-    ``list_dir(P + "/")``. Any returned prefix not being ``c`` or
-    starting with ``__`` indicates a child group implied by some
-    descendant group or array.
+    ``list_dir(P + "/")``. Any returned prefix not starting with ``__``
+    indicates a child group implied by some descendant group or array.
```
rabernat, line 1458:
```suggestion
    ``list_dir(P + "/")``. Any returned prefix ``Q`` not starting with ``__``
    indicates a potential child array or group. To determine if the child is an array or group,
    the document ``meta_key(Q)`` must be checked.
```

What I dislike about this is that two sequential operations are required to discover children. On high latency stores, this creates significant overhead.

Does anyone see an alternative algorithm?

docs/v3/core/v3.0.rst
```diff
@@ -1601,6 +1590,9 @@ a prefix will erase all the implicit group in the prefix.
 Care must thus be taken when erasing an array or a group if the parent needs to
 be converted into an explicit group.
 
+A race-condition arises if a client writes an array at path ``P``,
+and another concurrently assumes ``P`` is an implicit group and writes subgroups or arrays into it.
```
rabernat, line 1594:
```suggestion
and another concurrently assumes ``P`` is an implicit group and writes subgroups or arrays into it.
Implementations may choose to never use implicit groups to avoid this.
```

rabernat (commented):
Looking good. A few comments.

docs/v3/core/v3.0.rst
jbms: I think "gs://", "s3://", "https://", etc.

docs/v3/core/v3.0.rst
jbms: I think we can assume that all members are in fact children.  If we want to know if a given child is an array or a group, we do have to read the zarr.json metadata, and that is unavoidable.  As we discussed previously, though, it is likely that in cases where you want to know if it is an array or a group, you will also want to know additional details that would require reading the zarr.json metadata anyway.

docs/v3/stores/filesystem/v1.0.rst
jbms: Ah okay, I didn't realize that an extra initial slash was used even on Windows.

docs/v3/core/v3.0.rst
jstriebel: Exactly. Should we add some clarification in the spec, @rabernat?

docs/v3/core/v3.0.rst
jstriebel: I rephrased this slightly to clarify it:
> To discover the children of a group at hierarchy path `P`, perform
    ``list_dir(P + "/")``. Any returned prefix ``Q`` not starting with ``__``
    indicates a child array or group. To determine whether the child is
    an array or group, the document ``meta_key(Q)`` must be checked.
    
Reading the metadata is only needed when one needs to clarify if it's a group or array.

jstriebel:
@rabernat @jbms Is anything missing from your side?

jbms (approved):

PR #214, opened 2023-02-23
Add chunk_key_encoding array metadata field

### LINKED ISSUE #172: Chunk key encoding
There are a few issues with the encoding of chunk (and shard) keys that I'd like to raise.

With the current v3 spec, a chunk at position `(x, y, z)` is encoded as `cx/y/z` by default.  The same naming convention is also used for shards under the sharding extension.

The only customization currently supported is via the `chunk_grid/separator` metadata value.  The current specification doesn't describe any requirements on that field; it just provides a few examples.

At the very least, we should properly specify that field, and should constrain it to either "/" or "." as in the zarr v2 specificatio. 
 As discussed in the context of the zarr v2 `dimension_separator` field, allowing arbitrary separators can lead to ambiguities and there aren't any known use cases.

However, there are a few other issues with this encoding:

In many cases, the underlying store will be ordered and support range queries on keys, meaning listing all the keys lexicographically >= a start key and lexicographically < an end key.  For example, S3, GCS, and most ordered database tables.  More sophisticated stores may support additional operations, like efficiently querying the total number of bytes stored within a key range, without enumerating all of the keys.  If we choose the encoding of zarr chunks as keys appropriately, we can take advantage of this operation to efficiently determine whether a given subregion of the array contains any data (or possibly how much data).  

When processing a large array that is sparsely written (i.e. not all chunks are present), this type of query allows us to efficiently skip empty regions of the array.

Unfortunately, the current key encoding in the zarr v3 spec is not optimal for use with range queries:
1. Lexicographical order does not match numeric order: "c10" is lexicographically less than "c9"
2. While we support both C and F order for encoding individual chunks, the key encoding always corresponds (aside from point 1) to C order.  In most cases, it would probably be desirable for the key encoding to match the chunk encoding, but in general the user should be able to choose any order.

Regarding point 1, I think there was a comment on ZEP 1 to use zero padding when encoding chunk keys, but I'm having trouble finding it.  The drawback to zero padding is that we have to choose a fixed amount of zero padding, which may not be sufficient if we resize the array.  We could always use the maximum possible amount of zero padding (assuming array bounds are limited to 64-bit integer) but then the keys may be annoyingly long.  The advantage to zero padding is that it remains extremely readable.

There are variable-length encodings of numbers where lexicographical order matches numeric order.  For example, such a scheme is used by UTF-8.  A simpler alternative might be to prefix each component with the number of digits, in base 16.  E.g. instead of "c10" and "c9" we would have "210" and "19".

It is indeed fairly straightforward to define extensions that change the key encoding, so arguably we could defer this issue until later.  However, the current key encoding in the spec is not optimal for anything, except perhaps human readability, so it may be worthwhile to improve it even if we defer some key encoding improvements (e.g. morton order) to extensions.

jbms:
Fixes #172.


docs/v3/core/v3.0.rst
```diff
@@ -1042,8 +1086,13 @@ chunking as above, but using an extension data type::
         "chunk_grid": {
             "name": "regular",
             "configuration": {
-              "chunk_shape": [1000, 100],
-              "separator" : "/"
+              "chunk_shape": [1000, 100]
```
jstriebel, line 1089: Just noticed this since you fixed it already above:

```suggestion
                "chunk_shape": [1000, 100]
```

docs/v3/core/v3.0.rst
```diff
@@ -647,18 +647,6 @@ that chunk, where "%" is the modulo operator. For example, if a
...
-
-- the ``separator`` character specified within the ``chunk_grid`` metadata object (see
-  the section on `Array metadata`_ below), followed by,
-
-- the ASCII decimal string representation of the chunk index within that dimension.
-
-For example, in a 3 dimensional array, with a separator of ``/`` the identifier
-for the chunk at grid index (1, 23, 45) is the string "c/1/23/45".  With a
-separator of ``.``, the identifier is the string "c.1.23.45".
-
```
jstriebel, line 661: Please add a reference to the chunk_key_encoding metadata section here.

jstriebel (commented):
Thanks  @jbms, mostly LGTM! Just two minor comments:

docs/v3/core/v3.0.rst
```diff
@@ -895,6 +875,65 @@ mandatory names:
...
+         changed from ``.`` to ``/``, as in N5.  This decreases the maximum number of
+         items in hierarchical stores like directory stores.
+
+      .. note:: Arrays may have 0 dimensions (when for example representing scalars),
+         in which case the coordinate of a chunk is the empty tuple, and the chunk key
+         will consist of the string ``c``.
+
+    - ``v2``
+
+      The ``configuration`` object may contain one optional member,
```
jstriebel, line 918: Maybe add a short hint that this is purely meant for backwards-compatibility (e.g. only writing new metadata).

jbms: Done.

docs/v3/core/v3.0.rst
jbms: Done.

docs/v3/core/v3.0.rst
jbms: Done.

jstriebel (approved):
LGTM, thanks!

jstriebel:
My approval doesn't count since I pushed the conflict resolution. @jbms It might even work if you approve it yourself. cc @joshmoore @ryan-williams Could one of you approve this PR in case this doesn't work? Thanks :-)

joshmoore (approved):
Approving for @jstriebel

PR #218, opened 2023-02-28
Fix data_type example

normanrz:


jstriebel (commented):
Thanks! :+1: Could you also add similar changes in the two other places where the outdated dtype is used, L1005 and L1040? Then my approval counts for mergability, if I add changes myself someone else must approve it, too.

jstriebel (approved):

Issue #220, opened 2023-03-09
sharding as a codec rather than array storage transformer


jbms:
In the ZEP meeting on 2023-03-02, @jstriebel, @normanrz and I discussed representing sharding (ZEP 2) in the metadata as a codec rather than a storage transformer:

```json
{"chunk_grid": {
  "name": "regular",
  "configuration": {"chunk_shape": [1000, 1000, 1000]}},
  "codecs": [{"name": "sharding_indexed", "configuration": {
    "chunk_grid": {"name": "regular", "configuration": [100, 100, 100]}, 
    "codecs": [{"name": "blosc"}]
  }}],
 "shape": [50000, 50000, 50000],
 "data_type": "uint16",
 "zarr_format": 3,
 "node_type": "array",
 "fill_value": 0
}
```

In general, the idea here is to specify the partitioning of the array top-down, i.e. the `chunk_grid` specifies the top-level partitioning, and then the codec is responsible for any further subdivision, while with the previous proposal to represent sharding with a storage transformer, we are specifying the partitioning bottom-up, i.e. the `grid_grid` specifies the bottom-level partitioning, and then the storage transformer builds up larger shards from the individual chunks.  

While the sharding format as proposed in ZEP2 works fine as a storage transformer, in terms of the spec there are some advantages to making it a codec:

- Previously sharding had to "hijack" `chunk_key_encoding`, which should be specifying the keys for chunks, to instead specify the keys for shards; at the same time, the `chunk_key_encoding` was still used to create fake chunk keys for the individual chunks that served only as an identifier to pass around internally within the implementation, since the sharding format does not store keys.  With sharding as a codec, each shard is in fact a chunk and there is no need for string keys for the sub-chunks. 

The greater benefit to representing sharding as a codec is in enabling some interesting future extensions beyond the sharding format of ZEP2:

- The NVIDIA nvcomp project is interested in a sharding format where the compression is specified within the shard format, and may differ across shards.  With sharding as a codec, this is straightforward and very natural.  With sharding as a storage transformer, this can be done but basically requires that we leave the `codecs` empty in the array metadata.

- It is easy to imagine a sharding format where the sub-chunking of each shard is specified as part of a header/footer in the shard itself, and may differ across shards.  That is impossible to support with storage transformers, because the bottom-level chunking may vary across the different referenced arrays.

There are a few things that are possible with storage transformers but are not possible with a codec:
- Shards corresponding to non-rectangular regions of the array.
[...]

PR #222, opened 2023-03-16
Describe partial decode support for codecs

jbms:
This is a draft of the spec change to resolve issue https://github.com/zarr-developers/zarr-specs/issues/220.

### DIFF
--- a/docs/v3/codecs/transpose/v1.0.rst
+++ b/docs/v3/codecs/transpose/v1.0.rst
@@ -72,7 +72,7 @@ order:
 Format and algorithm
 ====================
 
-This is an ``array -> array` codec.
+This is an ``array -> array`` codec.
 
 Given a chunk array ``A`` with shape ``A_shape`` as the decoded representation,
 the encoded representation is an array ``B`` with the same data type as ``A``
--- a/docs/v3/codecs/transpose/v1.0.rst
+++ b/docs/v3/codecs/transpose/v1.0.rst
@@ -26,7 +26,8 @@ is licensed under a `Creative Commons Attribution 3.0 Unported License
 Abstract
 ========
 
-Defines a codec that permutes the dimensions of the chunk array.
+Defines an ``array -> array`` codec that permutes the dimensions of the chunk
+array.
 
 
 Status of this document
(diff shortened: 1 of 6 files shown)


normanrz (commented):
Looks pretty good. I would note for each registered codec whether it is {bytes,array} -> {bytes,array}. Also, we could add to blosc and sharding codecs that partial decoding is recommended for implementations.

jstriebel:
I agree with @normanrz, every codex should declare what its inputs and outputs are. Apart from that LGTM 👍

jbms:
I rebased this PR, and added a description for each codec of what type of codec it is.

docs/v3/codecs/transpose/v1.0.rst
```diff
@@ -71,9 +72,7 @@ order:
 Format and algorithm
 ====================
 
-The decoded chunk representation to which this codec is applied must be an
-array.  Implementations must fail if this codec is specified immediately after
-another codec that produces a byte string as its encoded representation.
+This is an ``array -> array` codec.
```
normanrz, line 75:
```suggestion
This is an ``array -> array`` codec.
```

docs/v3/core/v3.0.rst
```diff
@@ -1089,6 +1089,36 @@ each of these two representations are defined to be either:
...
+the following form:
+
+- zero or more ``array -> array`` codecs; followed by
+- at most one ``array -> bytes`` codec; followed by
+- zero or more ``bytes -> bytes`` codecs.
+
+If no ``array -> bytes`` codec is specified, then the default byte
+representation for the data type of the array is used.  For all data types
+currently defined by the core spec, that is equivalent to the ``endian`` codec
+with an endianness of ``little``.
```
normanrz, line 1120: Should we also mention the `transpose` codec with order `C`?

jbms: The endian codec with endianness of little may be a noop in terms of the actual in-memory representation of the data, but it does have an effect in terms of the logical data model, in that it transforms a multidimensional array into a (one-dimensional) byte sequence.  In contrast, the transpose codec with order [0, 1, ..., rank-1], i.e. "C", is logically an identity transform.  It is the endian codec that uses C order, and if you want a different order you have to transpose it first.

normanrz (approved):

jstriebel (approved):
LGTM 👍

PR #223, opened 2023-03-23
Rewrite sharding as a codec

normanrz:
This PR rewrites the sharding specification from ZEP2 as a codec.


docs/v3/codecs/sharding-indexed/v1.0.rst
```diff
@@ -0,0 +1,224 @@
...
+Sharding logically splits chunks ("shards") into sub-chunks ("inner chunks") 
+that can be individually compressed and accessed. This allows to colocate 
+multiple chunks within one storage object, bundling them in shards.
+
+
+Motivation
+==========
+
+In many cases, it becomes inefficient or impractical to store a large number of
+chunks in single files or objects due to the design constraints of the
```
jbms, line 42: "in single files" -> "as separate files"

docs/v3/codecs/sharding-indexed/v1.0.rst
```diff
@@ -0,0 +1,224 @@
...
+underlying storage. For example, the file block size and maximum inode number
+restrict the usage of numerous small files for typical file systems, also cloud
+storage such as S3, GCS, and various distributed filesystems do not efficiently
+handle large numbers of small files or objects.
+
+Increasing the chunk size works only up to a certain point, as chunk sizes need
+to be small for read and write efficiency requirements, for example to stream
```
jbms, line 49: since this solution does not address the write efficiency issue of large chunks, perhaps leave that out

docs/v3/codecs/sharding-indexed/v1.0.rst
```diff
@@ -0,0 +1,224 @@
...
+                  "level": 1
+                }
+              }
+            ]
+          }
+        }
+      ]
+    }
+
+``chunk_shape``
```
jbms, line 105: Rather than `chunk_shape` here, it would be nice to use `chunk_grid` exactly as in the top-level array metadata.  That way this can accommodate a rectilinear chunk grid for the inner chunks as well.  However, I realized there is a complication:

Ideally we could support a fully general rectilinear sub-chunk grid, e.g.:


```json
{
  "chunk_grid": {
    "type": "rectilinear",
    "configuration": {
      "chunk_shape": [[15, 35]]
    }
  },
  "codecs": [{
    "name": "sharding_indexed",
    "configuration": {
      "chunk_grid": {
        "type": "rectilinear",
        "chunk_shape": [[5, 10, 15, 20]]
      },
      "codecs": [{"name": "gzip"}]
    }
  }]
}
```

Here, the `chunk_grid` specified for `sharding_indexed` applies to the overall array, rather than the individual chunk.  The chunk boundaries must be aligned with the top-level chunk boundaries.  This allows us to have a different sub-chunk grid for each shard, which provides the most flexibility.

However, the metadata structure would make it more natural for the `chunk_grid` specified for `sharding_indexed` to apply only to the shard itself, which would mean the same chunk grid has to be used for every shard, and means that rectilinear chunks don't really interoperate with sharding well.

There is also another problem with having the `chunk_grid` specified for `sharding_indexed` apply to the overall array: it requires that the codec know which region of the overall array it is operating on.  But if the `transpose` codec, for example, is applied before `sharding_indexed`, then it is not clear how to provide `sharding_indexed` with information about its position in the overall array.

normanrz: I am happy to change the metadata to the `chunk_grid` object. Since there isn't a specific proposal for rectilinear arrays yet, I think we don't need to decide this yet. For now, I would add that the inner `chunk_grid` needs to be `regular`.

> However, the metadata structure would make it more natural for the `chunk_grid` specified for `sharding_indexed` to apply only to the shard itself, which would mean the same chunk grid has to be used for every shard, and means that rectilinear chunks don't really interoperate with sharding well.

It almost seems like the `chunk_grid` configuration for rectilinear grids would need to be added to the shard index.

jbms: I think you have a good point --- in practice if using rectilinear chunk grids, it would be most useful to be able to specify the chunking of each shard as part of the shard, rather than in the top-level metadata.  That would require a different shard format, though.

Using the same non-regular rectilinear chunking for every shard is unlikely to be useful, but specifying separate rectilinear chunking for each shard in the top-level metadata would be rather awkward and doesn't fit well with how codecs are supposed to work.

Therefore, I withdraw this suggestion; I think it is fine with just a `chunk_shape` member as you have it, rather than using `chunk_grid`.

normanrz: I actually think keeping the `chunk_grid` in is a good idea for later extensibility and consistent with the v3 mission. If we would adding support for rectilinear grids through the binary format, it would need to be specified somewhere. Also, we might come up with other grids that make more sense for sharding.

jbms: At least the current rectilinear grid proposal puts all of the partition points in the array metadata itself, e.g. `{"name": "rectilinear", "chunk_shape": [[100, 100, 50, 100], [150, 150, 25, 100], [200, 300, 100, 200]]}`.  Whereas for sharding we would perhaps just specify `{"name": "rectilinear"}` but then rely on a quite different binary format to actually specify the grid.    Additionally, the type of grid could also be specified in the binary format, as you could imagine that even with a regular grid, you might want different shards to use a different chunk shape.  Given these differences I'm not sure it makes sense to try to tie the shard chunking to the top-level array chunking.  Instead there could just be a different codec name.

normanrz: Fine with me.

normanrz:
@jstriebel Would you mind having a look at these changes?
https://zarr-specs--223.org.readthedocs.build/en/223/v3/codecs/sharding-indexed/v1.0.html

docs/v3/codecs/sharding-indexed/v1.0.rst
```diff
@@ -0,0 +1,220 @@
...
+
+    An array of integers providing the shape of inner chunks in a shard for each 
+    dimension of the Zarr array. The length of the array must match the length
+    of the array metadata ``shape`` entry. The each integer must by divisible by 
+    the ``chunk_shape`` of the array as defined in the ``chunk_grid`` 
+    :ref:`_array-metadata`.
```
jstriebel, line 107: This link does not work atm.

docs/v3/codecs/sharding-indexed/v1.0.rst
```diff
@@ -0,0 +1,220 @@
...
+    For example, an inner chunk shape of ``[32, 2]`` with an outer chunk shape 
+    ``[64, 64]`` indicates that 64 chunks are combined in one shard, 2 along the 
+    first dimension, and for each of those 32 along the second dimension.
+    Currently, only the ``regular`` chunk grid is supported.
+
+``codecs``
+
+    Specifies a list of codecs to be used for encoding and decoding inner chunks. 
+    The value must be an array of objects, as specified in the 
+    :ref:`_array-metadata`. An absent ``codecs`` member is equivalent to
```
jstriebel, line 117: Same here.

jstriebel (approved):
Nice, mostly LGTM :+1: Just found two links which are not working.

normanrz:
Thanks! Is there a tool to check the links?

jstriebel:
> Thanks! Is there a tool to check the links?

I think it's in the output of sphinx, but you might need to remove the cache to see all warnings.

jstriebel (approved):
LGTM! Should we merge this together with https://github.com/zarr-developers/zeps/pull/34?

normanrz:
> LGTM! Should we merge this together with [zarr-developers/zeps#34](https://github.com/zarr-developers/zeps/pull/34)?

Yes, please!

jbms (approved):

PR #225, opened 2023-03-24
Revise how the Blosc codec is specified

jbms:
The `shuffle` mode is now specified more clearly as `null` (0 in Zarr v2), `"bit"` (2 in Zarr v2), or `"byte"` (1 in Zarr v2).  Using these constants rather than numbers makes it much easier to know what shuffle mode will be using from manual inspection of the metadata.

When shuffling is enabled, the `typesize` must now be specified explicitly in the metadata, rather than determined implicitly from the input data.  This allows Blosc to function as a pure "bytes -> bytes" codec rather than an "array -> bytes" codec.

The special `shuffle` value of `-1`, which indicated to choose automatically between bit-wise or byte-wise shuffling depending on the typesize value, is not supported in the metadata.

### DIFF
--- a/docs/v3/codecs/blosc/v1.0.rst
+++ b/docs/v3/codecs/blosc/v1.0.rst
@@ -98,7 +98,7 @@ size::
             "configuration": {
                 "cname": "lz4",
                 "clevel": 1,
-                "shuffle": "byte",
+                "shuffle": "shuffle",
                 "typesize": 4,
                 "blocksize": 0
             }
@@ -123,8 +123,8 @@ Comparison to Zarr v2
 While the binary format is identical, the JSON metadata differs from that used
 by the Zarr v2 ``blosc`` codec in the following ways:
 
-- The `shuffle` mode is now specified more clearly as `null` (0 in Zarr v2),
-  `"bit"` (2 in Zarr v2), or `"byte"` (1 in Zarr v2).  Using these constants
+- The `shuffle` mode is now specified more clearly as `noshuffle` (0 in Zarr v2),
+  `"bitshuffle"` (2 in Zarr v2), or `"shuffle"` (1 in Zarr v2).  Using these constants
   rather than numbers makes it much easier to know what shuffle mode will be
   using from manual inspection of the metadata.
 
--- a/docs/v3/codecs/blosc/v1.0.rst
+++ b/docs/v3/codecs/blosc/v1.0.rst
@@ -70,25 +70,27 @@ clevel:
     level is 0.
 
 shuffle:
-    An integer value in the set {0, 1, 2, -1} indicating the way
-    bytes or bits are rearranged, which can lead to faster
-    and/or greater compression. A value of 1
-    indicates that byte-wise shuffling is performed prior to
-    compression. A value of 2 indicates the bit-wise shuffling is
-    performed prior to compression. If a value of -1 is given,
-    then default shuffling is used: bit-wise shuffling for buffers
-    with item size of 1 byte, byte-wise shuffling otherwise.
-    Shuffling is turned off completely when the value is 0.
+    Specifies the type of shuffling to perform, if any, prior to compression.
+    Must be one of:
+
+    - ``"noshuffle"``, to indicate no shuffling;
+    - ``"shuffle"``, to indicate byte-wise shuffling;
+    - ``"bitshuffle"``, to indicate bit-wise shuffling.
+
+typesize:
+    Positive integer specifying the stride in bytes over which shuffling is
+    performed.  Required unless ``shuffle`` is ``"noshuffle"``, in which case the value
+    is ignored.
 
 blocksize:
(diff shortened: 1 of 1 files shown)


normanrz:
cc @jakirkham

docs/v3/codecs/blosc/v1.0.rst
```diff
@@ -70,33 +70,36 @@ clevel:
...
-    performed prior to compression. If a value of -1 is given,
-    then default shuffling is used: bit-wise shuffling for buffers
-    with item size of 1 byte, byte-wise shuffling otherwise.
-    Shuffling is turned off completely when the value is 0.
+    Specifies the type of shuffling to perform, if any, prior to compression.
+    Must be one of:
+
+    - ``null``, to indicate no shuffling;
+    - ``"byte"``, to indicate byte-wise shuffling;
+    - ``"bit"``, to indicate bit-wise shuffling.
```
jstriebel, line 78: I'm in favor of those changes :+1:

docs/v3/codecs/blosc/v1.0.rst
```diff
@@ -70,33 +70,36 @@ clevel:
...
+
+typesize:
+    Positive integer specifying the stride in bytes over which shuffling is
+    performed.  Required unless ``shuffle`` is ``null``, in which case the value
+    is ignored.
```
jstriebel, line 83: I'm not sure if this change is useful atm. I have a two of questions:
- Couldn't we make this optional, and have a auto-detect logic as before? It would only be necessary for dtypes with arbitrary length data.
- Why should it be called `typesize`? It could also be any valid divisor of the typesize, especially for arbitrary-length types, right?

I'd rather leave this out for now, and add it later once we have dtypes with arbitrary sizes. Then this can be an additional optional field. Although this would be nice:
> This allows Blosc to function as a pure "bytes -> bytes" codec rather than an "array -> bytes" codec.

Not quite sure if we should include this. cc @joshmoore @normanrz @rabernat @jakirkham What do you think about it?

docs/v3/codecs/blosc/v1.0.rst
jbms: Looking at the blosc API more now, I think these should probably be changed to be consistent with what the blosc API uses:

"noshuffle", "shuffle", "bitshuffle"

docs/v3/codecs/blosc/v1.0.rst
jstriebel: > If this is preferred I can try to write up the necessary changes to core specification.

I'd be in favor of this (Option 1), but would like to have some more opinions first. In my opinion this is blocking ZEP 1 / v3.

normanrz: No strong opinion here. Option 1 is fine with me. So is keeping `typesize` in. It only seems like a small redundancy. Might facilitate fallback dtype handling.

jbms:
I updated the shuffle constants to "noshuffle", "shuffle" and "bitshuffle" to match the blosc C API.  Per discussions in the ZEP1 review and the recent ZEP meeting, I have retained typesize.

docs/v3/codecs/blosc/v1.0.rst
```diff
@@ -114,6 +117,31 @@ reference implementation is provided by the `c-blosc library
...
 
 
+Comparison to Zarr v2
+=====================
+
+While the binary format is identical, the JSON metadata differs from that used
+by the Zarr v2 ``blosc`` codec in the following ways:
+
+- The `shuffle` mode is now specified more clearly as `null` (0 in Zarr v2),
+  `"bit"` (2 in Zarr v2), or `"byte"` (1 in Zarr v2).  Using these constants
```
jstriebel, line 127:
```suggestion
- The `shuffle` mode is now specified more clearly as `noshuffle` (0 in Zarr v2),
  `"bitshuffle"` (2 in Zarr v2), or `"shuffle"` (1 in Zarr v2).  Using these constants
```

docs/v3/codecs/blosc/v1.0.rst
```diff
@@ -114,6 +117,31 @@ reference implementation is provided by the `c-blosc library
...
+  "array -> bytes" codec.  Zarr implementations MAY allow users to leave this
+  unspecified and have the implementation choose a value automatically based on
+  the array data type and previous codecs in the chain, but MUST record in the
+  metadata the value that is chosen.
+
+- There is no option to choose between bit-wise and byte-wise shuffling
+  automatically, as supported in Zarr v2 via a `shuffle` value of `-1`.  Zarr
+  implementations MAY provide users an option to choose a shuffle mode
+  automatically based on the typesize or other information, but MUST record in
+  the metadata the mode that is chosen.
```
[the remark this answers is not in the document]
jbms: Done.

jstriebel (approved):
🎉

PR #228, opened 2023-04-07
[WIP] styling

d-v-b:
I tightened up some of the prose early in the document and fixed a few omissions. I'm not quite done.  When I finish, I will remove the WIP flag.

### DIFF
--- a/docs/v3/core/v3.0.rst
+++ b/docs/v3/core/v3.0.rst
@@ -55,35 +55,34 @@ Introduction
 
 This specification defines a format for multidimensional array data. This
 type of data is common in scientific and numerical computing
-applications. Many domains are facing computational challenges as
+applications. Many domains face computational challenges as
 increasingly large volumes of data are being generated, for example,
 via high resolution microscopy, remote sensing imagery, genome
 sequencing or numerical simulation. The primary motivation for the
-development of Zarr has been to help address this challenge by
+development of Zarr is to address this challenge by
 enabling the storage of large multidimensional arrays in a way that is
 compatible with parallel and/or distributed computing applications.
 
-This specification is intended to supersede the `Zarr storage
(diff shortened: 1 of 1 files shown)


docs/v3/core/v3.0.rst
```diff
@@ -1556,7 +1555,7 @@ Let "+" be the string concatenation operator.
 Storage transformers
 ====================
 
-A Zarr storage transformer allows to change the zarr-compatible data before storing it.
+A Zarr storage transformer allows a store to change zarr-compatible data before storing it.
```
jbms, line 1558: I am unsure about this change --- not sure if we consider a "storage transformer" a "store".  Everything else here looks good, though.

d-v-b: i wasn't sure about this one either. Perhaps instead of "A Zarr storage transformer allows a Y to do X"  we can just say "A Zarr storage transformer does X", e.g. "A Zarr storage transformer modifies a request to store or retrieve data before passing that request to a store"

jstriebel: > "A Zarr storage transformer modifies a request to read or write data before passing that request to a store"

This still refers to the next entitity after the transformer as a "store", but it could also be a storage transformer. So maybe
```suggestion
"A Zarr storage transformer modifies a request to read or write data before passing that request."
```
or
```suggestion
"A Zarr storage transformer modifies a request to read or write data before passing that request to the following transformer or store."
```

jbms:
@d-v-b is this ready to merge?  If you have other changes you could open a separate PR.

jstriebel (approved):
LGTM :+1:

My approval doesn't count anymore since I added commits myself. @jbms @rabernat, @joshmoore, can one of you approve, too?

PR #230, opened 2023-04-30
minor grammar/typo fixes to v3 spec

grlee77:
The v3 spec is substantial clearer than when I last read it ~ 1 year ago. Congratulations to all who have worked to improve it over that time. I especially thank getting rid of split data/ folder and the `root/` requirement from paths makes things substantially easier on the implementation side!

This MR just proposes a few minor issues I noticed when reading it. It does not propose any changes in behavior.

### DIFF
--- a/docs/v3/core/v3.0.rst
+++ b/docs/v3/core/v3.0.rst
@@ -131,7 +131,7 @@ implementing a specification ``X.Y`` can be considered compatible with all
 datasets which only use features contained in version ``X.Y``.
 
 For example, spec ``X.1`` adds core feature "foo" compared to ``X.0``. Assuming
-implementation A implements ``X.0`` and implementation B implements ``X.1``.
+implementation A implements ``X.1`` and implementation B implements ``X.0``.
 Data using feature "foo" can only be read with implementation A. B fails to open
 it, as the key "foo" is unknown.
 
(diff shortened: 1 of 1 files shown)


grlee77:
Two other minor comments/questions for clarification:

1.) Under the `get_partial_values` method description it states:
> Output: list of `values`, in the order of the `key_ranges`, may contain none for missing keys

If the user is providing the (key, range) tuples as input, under what scenario(s) are there "missing keys"? I thought the returned values should always correspond to the provided keys in the order given?

2.) The abstract store interface describes keys as needing to be in the format
> a-z``, ``A-Z``, ``0-9``, or in the set ``/.-_``, where the final character is **not** a ``/`` character

However, under node names, this is the recommended character set, but it is not a requirement. Should the "node name" and "store key" naming restrictions be the same?

jbms (approved):

jbms:
> Two other minor comments/questions for clarification:
> 
> 1.) Under the `get_partial_values` method description it states:
> > Output: list of `values`, in the order of the `key_ranges`, may contain none for missing keys
> 
> If the user is providing the (key, range) tuples as input, under what scenario(s) are there "missing keys"? I thought the returned values should always correspond to the provided keys in the order given?

If the key is not present in the store (e.g. the file does not exist) that is considered a missing key.

> 
> 2.) The abstract store interface describes keys as needing to be in the format
> > a-z``, ``A-Z``, ``0-9``, or in the set ``/.-_``, where the final character is **not** a ``/`` character

Yes, that is out of date.  It should now be any unicode string subject to store-specific limitations.

> 
> However, under node names, this is the recommended character set, but it is not a requirement. Should the "node name" and "store key" naming restrictions be the same?
>

joshmoore:
Merging so other reviewers don't hit upon the same minor issues but :+1: for keeping the conversation going :smile:

PR #233, opened 2023-05-06
Apply suggestions from #230

jstriebel:
This tries to improve the points @grlee77 raised in #230:
* Corrects the requirements of key in the abstract store interface section.
* Clarifies what happens to missing keys in `get_partial_values`.

### DIFF
--- a/docs/v3/core/v3.0.rst
+++ b/docs/v3/core/v3.0.rst
@@ -1303,9 +1303,9 @@ that an implementation of this specification could be modular and
 allow for different store implementations to be used.
 
 The store interface defines a set of operations involving `keys` and
-`values`. In the context of this interface, a `key` is any string
-containing only characters in the ranges ``a-z``, ``A-Z``, ``0-9``, or
-in the set ``/.-_``, where the final character is **not** a ``/``
+`values`. In the context of this interface, a `key` is a string, more
+precisely a concatenation of the `node_names`_ using ``/`` as the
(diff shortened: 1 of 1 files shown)


docs/v3/core/v3.0.rst
```diff
@@ -1303,9 +1304,9 @@ that an implementation of this specification could be modular and
 allow for different store implementations to be used.
 
 The store interface defines a set of operations involving `keys` and
-`values`. In the context of this interface, a `key` is any string
-containing only characters in the ranges ``a-z``, ``A-Z``, ``0-9``, or
-in the set ``/.-_``, where the final character is **not** a ``/``
+`values`. In the context of this interface, a `key` is a string, more
```
jbms, line 1307: I would say a "Unicode string" to clarify that it is not a byte string.

Also I'm not sure that we should reference node names here, since for example we exclude "__" prefix for node names but that does not apply to the keys here.

jstriebel: > I would say a "Unicode string" to clarify that it is not a byte string.

👍

> Also I'm not sure that we should reference node names here, since for example we exclude "__" prefix for node names but that does not apply to the keys here.

Wouldn't this then still hold for the parts of the key? I agree that it's probably not relevant or helpful in this section, but a link seems fine IMO.

jstriebel: I agree, the link is not useful here.

jbms (approved):

Issue #231, opened 2023-05-05
Zarr V3: questions about fill_value representation as hexadecimal representation and complex data types


rouault:
Reading https://zarr-specs.readthedocs.io/en/latest/v3/core/v3.0.html#fill-value, I'm not sure how the nodata value for complex data types should be encoded. Should it be an array of two elements like ``[1.5, 2.5]`` ? Is an hexadecimal representation allowed, and if so how should it be encoded ?

Regarding the hexadecimal representation, it is a bit ambiguous for Float data types. I assume this should be the integer corresponding to the IEEE754 encoding of the floating point number, and that for a Float32 it should be 0xABCDEFGH where AB is the most significant byte ?

jstriebel:
Agreed, this is currently both missing. @rouault or @jbms, would one of you submit a PR?

rouault:
> would one of you submit a PR?

Probably a bit beyond what I can properly write in proper English in a formal doc :-) I'm not clear either how you would format "two consecutive little endian IEEE754 binary64 values" in a string.  Probably would require some separator, like "0xEFCDAB8967452301,0xEFCDAB8967452301" but that's going to be painful to parse, and I believe that requiring an array of strings or array of numbers for complex data is better.

And I'm not sure why "little endian" is mentionned. An hexadecimal representation usually uses a MSB convention (like the decimal writing), independently of the machine byte ordering. e.g "0xEFCDAB8967452301" represents the 64-bit word of least significant part 01 and most significant part 0xEF.
That's also how Python / C library formats numbers. For example the representation of 1.5 as the hexadecimal representation of the corresponding IEEE754 64-bit float would be:
>>> hex(struct.unpack('Q', struct.pack('d', 1.5))[0])
'0x3ff8000000000000'

jstriebel:
> "two consecutive little endian IEEE754 binary64 values" in a string

I assume that we would simply represent them as a JSON array in this case instead of a string.

jstriebel:
> And I'm not sure why "little endian" is mentionned. An hexadecimal representation usually uses a MSB convention (like the decimal writing), independently of the machine byte ordering. e.g "0xEFCDAB8967452301" represents the 64-bit word of least significant part 01 and most significant part 0xEF.

I think that only holds true for little-endian representations. A hexadecimal representation usually follows the memory layout, e.g. the first byte is represented left in the hexadecimal representation, and therefore depends if the float is interpreted as little-endian or big-endian.

[...]

PR #236, opened 2023-05-08
Clarify fill value representation, change nan representation

jbms:
Fixes https://github.com/zarr-developers/zarr-specs/issues/231.

### DIFF
--- a/docs/v3/core/v3.0.rst
+++ b/docs/v3/core/v3.0.rst
@@ -613,22 +613,16 @@ mandatory names:
 
       - A JSON string of the form:
 
-        - ``"inf"``, denoting positive infinity;
-        - ``"-inf"``, denoting negative infinity;
-        - ``"nan"`` or ``"-nan"``, equivalent to ``"nan(0x1)"`` and
-          ``"-nan(0x1)"`` respectively;
-        - ``"nan(0xYYYY)"`` or ``"-nan(0xYYYY)"``, specifying a NaN value with
-          positive or negative sign bit and payload indicated by ``YYYY``, where
-          ``YYYY`` is the base-16 representation of a *non-zero* integer
-          indicating the NaN payload.  The NaN payload ``YYYY``` specifies the
-          mantissa bits in the NaN representation in reverse order; that is, the
-          least significant bit of ``0xYYYY`` is the most significant bit of the
-          mantissa.  According to IEEE 754-2019, a quiet NaN has a payload with
-          a least significant bit of 1 (most significant bit of the mantissa
-          equal to 1), while a signaling NaN has a payload with a least
-          significant bit of 0 (most significant bit of the mantissa equal to
-          0).  This is consistent with x86 and ARM, but not consistent with MIPS
-          where the behavior may depend on a flag.
+        - ``"+Infinity"``, denoting positive infinity;
+        - ``"-Infinity"``, denoting negative infinity;
+        - ``"NaN"``, denoting thenot-a-number (NaN) value where the sign bit is
+          0 (positive), the most significant bit (MSB) of the mantissa is 1, and
+          all other bits of the mantissa are zero;
+        - ``"0xYYYYYYYY"``, specifying the byte representation of the floating
+          point number as an unsigned integer.  For example, for ``float32``,
+          ``"NaN"`` is equivalent to ``"0x7fc00000"``.  This representation is
+          the only way to specify a NaN value other than the specific NaN value
+          denoted by ``"NaN"``.
 
         .. warning::
 
@@ -646,8 +640,20 @@ mandatory names:
       indicates a complex number with real component of -inf and imaginary
       component of positive NaN with payload of ``0x7``.
 
+    Raw data types (``r<N>``)
+      An array of integers, with length equal to ``<N>``, where each integer is
+      in the range ``[0, 255]``.
+
     Extensions to the spec that define new data types must also define the JSON
     fill value representation.
+
+    .. note::
+
+       The ``fill_value`` metadata field is required, but Zarr implementations
+       may provide an interface for creating a new array with which users can
+       leave the fill value unspecified, in which case a default fill value for
+       the data type will be chosen.  However, the default fill value that is
+       chosen MUST be recorded in the metadata.
 
 The following members are optional:
 
--- a/docs/v3/core/v3.0.rst
+++ b/docs/v3/core/v3.0.rst
@@ -636,9 +636,9 @@ mandatory names:
       components respectively, where each component is specified as defined
       above for floating point number.
 
-      For example, ``[1, 2]`` indicates ``1 + 2i`` and ``["-inf", "nan(0x7)"]``
+      For example, ``[1, 2]`` indicates ``1 + 2i`` and ``["-Infinity", "NaN"]``
       indicates a complex number with real component of -inf and imaginary
-      component of positive NaN with payload of ``0x7``.
+      component of NaN.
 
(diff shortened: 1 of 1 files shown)


jbms:
This eliminates the special "0x..." and "0b..." fill value syntax for floating point numbers and instead adds a special syntax for specifying the NaN payload consistent with the C99 standard.

However, this does not yet address the "r*" raw types.

jbms:
Previously, the `fill_value` attribute was worded to allow the special "0x..." and "0b..." syntax seemingly for any data type. In fact, though, with the current data types, there were only two real use cases for that syntax: specifying a precise NaN representation, and specifying a value for the raw "r*" data types.  Allowing it for all data types creates a conflict though with the future introduction of a "unicode string" or "json value" data type, for which we would want to use a JSON string to indicate the actual value.

Currently, with the change made by this PR, the spec does not indicate how to specify the fill value for a raw "r*" data type.  Presumably for those data types the fill value would indeed need to be specified as a sequence of bytes, either as a hex string, or perhaps a base64 string, or a JSON array of integers 0-255.

It occurred to me, though, that the primary use case listed for the raw "r*" data types is as fallback data types for other fixed-size data types defined by extensions to/future versions of the spec.  The problem with the current fallback data type mechanism is that we don't have a fallback fill value, which leads to the following options:
1. If a fallback data type is specified, the fill value must be valid for both the primary data type and fallback data type. 
 For example, if we define `bfloat16` as a new data type and wanted to support a fallback of `uint16`, then `bfloat16` would need to represent its fill value as an unsigned integer.  Alternatively, if we wanted to support a fallback of `r2`, then `bfloat16` would need to represent its fill value as a raw byte sequence.  For consistency, we could just require all fill values to be specified as raw byte sequences, and eliminate support for the more "natural" JSON representations.
2. If a fallback data type is specified, a fallback fill value must also be specified.
3. When using a fallback data type, any operation that depends on the fill value, such as partially writing a chunk or reading a missing chunk, could just fail.

docs/v3/core/v3.0.rst
```diff
@@ -597,30 +597,57 @@ mandatory names:
...
-    ``"+Infinity"`` and ``"-Infinity"`` are also understood for floating point
-    data types.
+    IEEE 754 floating point numbers (``float{16,32,64}``)
+      The value may be either:
 
+      - A JSON number, that will be rounded to the nearest representable value.
+
+      - A JSON string of the form:
+
+        - ``"inf"``, denoting positive infinity;
```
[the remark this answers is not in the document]
jbms: To be clear, the proposed nan syntax as defined does not force a quiet nan, the number given in the parentheses fully specifies all the mantissa bits.  But indeed it cannot be parsed using strtof/strtod, since those functions use an implementation-defined mapping.

Negative zero can already be specified with regular json syntax, no need for a separate representation for that.

It seemed simpler to just have one way to represent each number so that the implementation does not have to choose between two different representations.

But I suppose if just specifying the full binary representation of the float is preferable, that is fine too.  Note that we would still need to precisely specify which NaN is denoted by "NaN".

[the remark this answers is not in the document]
rouault: > Negative zero can already be specified with regular json syntax, no need for a separate representation for that.

ah, you're right. My testing was incorrect. I used -0 instead of -0.0 ...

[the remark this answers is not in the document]
jstriebel: > It seemed simpler to just have one way to represent each number so that the implementation does not have to choose between two different representations.

Good point, having a single representation would be nice. But the special `nan(…)` syntax seems quite custom, and in most cases the simple `nan` would do, which also duplicates `nan(0x1)` then.

> But I suppose if just specifying the full binary representation of the float is preferable, that is fine too. Note that we would still need to precisely specify which NaN is denoted by "NaN".

I assume we will also need a binary representation for raw dtypes, and also for more custom dtypes in the future. Therefore I'd also use this for the specific nan representations. But it would be great if you could still include this as a detail, specifying how to represent a specific nan.

docs/v3/core/v3.0.rst
```diff
@@ -597,30 +597,57 @@ mandatory names:
...
-    the integer data types defined in this specification, then the
-    value must be a number with no fraction or exponent part and must
-    be within the range of the data type.
+    The permitted values depend on the data type:
 
-    For any data type, the ``fill_value`` is required. The literal
-    ``null`` is not permitted. The fill value needs to be defined
-    so that the data is independent of implementation details. Internally
-    implementations may provide a default ``fill_value``, but that must
-    be converted to a fixed value in the stored metadata.
```
jstriebel, line 610: I think it might be useful to keep this sentence, to distinguish between metadata and library API level.

jstriebel (commented):
Thanks a ton @jbms!

docs/v3/core/v3.0.rst
jbms: Done.

jbms:
I reverted back to the "+Infinity", "-Infinity", 'NaN" syntax used by zarr v2, and added a hex representation for floats in place of the "nan(0xYYYY)" syntax.  I used the "big endian" representation since that is by far the more common way to display the raw bytes of a float as a hexadecimal integer.

For the "rN" raw data types, I instead used an array of numbers, [0, 255].  But I could change this back to the "0x..." syntax if that is preferred.

docs/v3/core/v3.0.rst
```diff
@@ -717,7 +750,7 @@ compressed using gzip compression prior to storage::
                 "level": 1
             }
         }],
-        "fill_value": "NaN",
+        "fill_value": "nan(0x7)",
```
rouault, line 753:
```suggestion
        "fill_value": "NaN",
```

docs/v3/core/v3.0.rst
```diff
@@ -597,30 +597,63 @@ mandatory names:
...
+           definition.
+
+    Complex numbers (``complex{64,128}``)
+      The value must be a two-element array, specifying the real and imaginary
+      components respectively, where each component is specified as defined
+      above for floating point number.
+
+      For example, ``[1, 2]`` indicates ``1 + 2i`` and ``["-inf", "nan(0x7)"]``
+      indicates a complex number with real component of -inf and imaginary
+      component of positive NaN with payload of ``0x7``.
```
rouault, line 641:
```suggestion
      For example, ``[1, 2]`` indicates ``1 + 2i`` and ``["-Infinity", "NaN"]``
      indicates a complex number with real component of -inf and imaginary
      component of NaN.
```

docs/v3/core/v3.0.rst
```diff
@@ -597,30 +597,63 @@ mandatory names:
...
-    ``"+Infinity"`` and ``"-Infinity"`` are also understood for floating point
-    data types.
+    IEEE 754 floating point numbers (``float{16,32,64}``)
+      The value may be either:
 
+      - A JSON number, that will be rounded to the nearest representable value.
+
+      - A JSON string of the form:
+
+        - ``"+Infinity"``, denoting positive infinity;
```
rouault, line 616: Note that Zarr V2 uses "Infinity" without leading '+' . Is the discrepancy ZarrV2 vs ZarrV3 intended ?

jbms: That wasn't intentional --- I changed it back.

jbms:
@jstriebel Is there anything left to do?

jstriebel (approved):
LGTM, thanks a lot @jbms! And sorry for coming back to this so late.

PR #237, opened 2023-05-09
Add CRC-32C checksum to sharded index format

jbms:


jbms:
I also clarified the language a bit to account for the possibility of nested sharding.

jbms:
@normanz Your input would be much appreciated.

normanrz (approved):
LGTM :+1:

clbarnes:
Worth combining this with a spec for a CRC-32C bytes->bytes codec to apply to chunks too? Seems like low-hanging fruit if we're pulling in that functionality anyway.

jbms:
I agree it makes sense to add a crc32c bytes->bytes codec, but I would just assume address that in a separate PR.

jstriebel (approved):
LGTM, thanks :+1:

PR #238, opened 2023-05-10
Remove "extensions" from example array metadata

clbarnes:
This key invalidates the metadata as it is neither specified as a core key, nor is it in object with `"must_understand": false`. It was already absent from the "made up extension data type" example and is not in the example group metadata either.


jbms (approved):
Thanks!

PR #248, opened 2023-06-29
Remove fallback data types from specification

jbms:
None of the completed/in-progress implementations have implemented fallback data type support, to my knowledge, and they can't really be implemented based on the current spec given that it is not specified how fill values are handled if a fallback data type is used.


jbms:
@normanz Please take a look.

normanrz (approved):

docs/v3/core/v3.0.rst
```diff
@@ -1797,6 +1781,7 @@ by time.
 Draft Changes
 --------------------------
 
+- Removed fallback data type support.
```
jstriebel, line 1784: Please add this into a separate section (e.g. "Provisionally Accepted") and also link the MR:

```suggestion
Changes after Provisional Acceptance
--------------------------

- Removed fallback data type support. `PR #248
  <https://github.com/zarr-developers/zarr-specs/pull/248>`_

Draft Changes
--------------------------

```

jstriebel (dismissed):
I still think this will be a useful addition once more data-type extensions exist, and we will rather need to figure out how to handle the fill value then. But fair to remove it for now. Mostly LGTM, just left one minor comment about the changelog entry.

jstriebel (approved):

PR #249, opened 2023-07-06
Require array -> bytes codec be specified in metadata

jbms:
This removes the concept of a "default binary representation". Instead, it is required that an "array -> bytes" codec be specified in the list of codecs.  Implementations are still free to choose a default "array -> bytes" codec when creating an array if not specified by the user.  This should slightly simplify implementations, and makes it clearer to humans when reading the metadata which "array -> bytes" codec will be used.


jbms:
@normanrz Please take a look.

normanrz (changes requested):
Please also mention the required array->bytes codec in https://github.com/jbms/zarr-specs/blob/remove-default-binary-representation/docs/v3/codecs/sharding-indexed/v1.0.rst?plain=1#L119-L122

jstriebel (commented):
Nice, this mostly LGTM. Please also add a changelog entry, preferably in a new section `Changes after Provisional Acceptance` (see https://github.com/zarr-developers/zarr-specs/pull/248#discussion_r1266875793).

jbms:
> Nice, this mostly LGTM. Please also add a changelog entry, preferably in a new section `Changes after Provisional Acceptance` (see [#248 (comment)](https://github.com/zarr-developers/zarr-specs/pull/248#discussion_r1266875793)).

Done.

docs/v3/core/v3.0.rst
```diff
@@ -1794,6 +1741,13 @@ All notable and possibly implementation-affecting changes to this specification
 are documented in this section, grouped by the specification status and ordered
 by time.
 
+Changes after Provisional Acceptance
+------------------------------------
+
+- It is now required to specify a ``bytes -> bytes`` codec in the ``codecs``
```
jstriebel, line 1747:
```suggestion
- It is now required to specify an ``array -> bytes`` codec in the ``codecs``
```

jstriebel (dismissed):

jstriebel (approved):

jstriebel (approved):

DennisHeimbigner:
I am still seriously confused about what is required to handle endianness.
My current interpretation is that the first entry in the codecs chain must be "bytes" and it does nothing
but pass thru the bytes unchanged.
However it is unclear if it is also intended that the "bytes" codec also can specify endianness,
or is there still a separate "endian" codec that must the second element in the chain and is mandatory
for arrays whose type size is > 1.
Also, when specifying the endianness as a codec, it must specify the type of the data being encoded.
Can that type differ from the type of the associated array? Admittedly this is probably not useful, but
how is that situation to be interpreted and which type  takes precedence: the array type or the endian codec type?

jbms:
The bytes codec specifies a conversion from the logical array representation to a specific physical.(byte sequence) representation.  Whether that just passes through the bytes or does some conversation depends on the specific in-memory representation used by the implementation.  For example, if the implementation stores bools packed as single bits in memory, they will need to be expanded to each occupy a whole byte.

There is no separate endian codec.

The data type is not specified as an option to the bytes codec, instead it is inferred from the data type specified in the metadata and the sequence of prior array to array codecs.  The endian configuration option is required if relevant for the data type, otherwise it is ignored.

DennisHeimbigner:
Two questions:
1. Where is endian-ness encoding performed?
2. Where is the numcodecs specification for the bytes codec?

jbms:
The endian conversion would presumably happen either in the encode/decode functions for the `bytes` codec, or not at all if the implementation has arranged to use the same endianness for the in-memory representation as in the stored representation.

I don't think zarr v3 support is implemented in zarr-python yet, but you can refer to some other implementations here:

https://github.com/google/tensorstore/blob/1b7417d4221c0e9fea50aef33e139fca1a439506/tensorstore/driver/zarr3/codec/bytes.cc#L192
https://github.com/google/neuroglancer/blob/master/src/neuroglancer/datasource/zarr/codec/bytes/decode.ts
https://github.com/scalableminds/zarrita/blob/78274781ad64aef95772eb4b083f7ea9b7d03d06/zarrita/codecs.py#L330

DennisHeimbigner:
I do not think that it is a good idea to have an essential (mandatory even) concept in the V3 spec
where the concept has no definition in the V3 spec.
Apparently the only definition is in python code, so it becomes necessary for non-python implementors to follow changes to that python code.
This makes life very difficult for non-python V3 implementations.

jbms:
When you say that there is an essential concept that is not defined in the V3 spec, which concept are you referring to?

DennisHeimbigner:
The "bytes" codec.

jbms:
The bytes codec is specified in this PR:
https://github.com/zarr-developers/zarr-specs/pull/263

DennisHeimbigner:
> The bytes codec is specified in this PR: #263

Thanks, that was the piece of information I was missing.

PR #253, opened 2023-07-19
Adds index_codecs to the sharding codec

normanrz:
This PR adds the `index_codecs` configuration to the sharding codec. It also adds the `crc32c` checksum bytes-to-bytes codec.

Index codecs have been proposed by @jbms: https://github.com/zarr-developers/zarr-specs/pull/152#issuecomment-1632689835
There is a prototype implementation in zarrita: https://github.com/scalableminds/zarrita/blob/codec-pipeline/zarrita/sharding.py#L507-L518

### DIFF
--- a/docs/v3/codecs/crc32c/v1.0.rst
+++ b/docs/v3/codecs/crc32c/v1.0.rst
@@ -1,0 +1,95 @@
+.. _crc32c-codec-v1:
+
+============================
+ CRC32C checksum codec (version 1.0)
+============================
+
+  **Editor's draft 26 July 2019**
+
+Specification URI:
+    https://zarr-specs.readthedocs.io/en/latest/v3/codecs/endian/v1.0.html
+Corresponding ZEP:
+    `ZEP 1 — Zarr specification version 3 <https://zarr.dev/zeps/draft/ZEP0001.html>`_
+Issue tracking:
+    `GitHub issues <https://github.com/zarr-developers/zarr-specs/labels/codec>`_
+Suggest an edit for this spec:
+    `GitHub editor <https://github.com/zarr-developers/zarr-specs/blob/main/docs/v3/codecs/endian/v1.0.rst>`_
+
+Copyright 2020 `Zarr core development team
+<https://github.com/orgs/zarr-developers/teams/core-devs>`_. This work
+is licensed under a `Creative Commons Attribution 3.0 Unported License
+<https://creativecommons.org/licenses/by/3.0/>`_.
+
+----
+
+
+Abstract
+========
+
+Defines an ``bytes -> bytes`` codec that appends a CRC32C checksum of the input bytestream.
+
+
+Status of this document
+=======================
+
+.. warning::
+    This document is a draft for review and subject to changes.
+    It will become final when the `Zarr Enhancement Proposal (ZEP) 2 <https://zarr.dev/zeps/draft/ZEP0002.html>`_
+    is approved via the `ZEP process <https://zarr.dev/zeps/active/ZEP0000.html>`_.
+
+
+Document conventions
+====================
+
+Conformance requirements are expressed with a combination of
+descriptive assertions and [RFC2119]_ terminology. The key words
+"MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in the normative
+parts of this document are to be interpreted as described in
+[RFC2119]_. However, for readability, these words do not appear in all
+uppercase letters in this specification.
+
+All of the text of this specification is normative except sections
+explicitly marked as non-normative, examples, and notes. Examples in
+this specification are introduced with the words "for example".
+
+
+Codec name
+==========
+
+The value of the ``name`` member in the codec object MUST be ``crc32c``.
+
+
+Configuration parameters
+========================
+
+None.
+
+
+Format and algorithm
+====================
+
+This is an ``bytes -> bytes`` codec.
+
+The codec computes the CRC32C checksum as defined in [RFC3720]_ of the input
+bytestream. The output bytestream is composed of the unchanged input byte 
+stream with the appended checksum. The checksum is represented as a 32-bit
+unsigned integer represented in little endian. 
+
+
+References
+==========
+
+.. [RFC2119] S. Bradner. Key words for use in RFCs to Indicate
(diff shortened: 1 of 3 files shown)


normanrz:
@jbms @jstriebel Please take a look. If we decide to go with this, it would be good to merge this before we send ZEP2 to the ZIC.

docs/v3/codecs/sharding-indexed/v1.0.rst
```diff
@@ -105,85 +120,104 @@ Sharding can be configured per array in the :ref:`array-metadata` as follows::
...
 
-Given the example of 2x2 inner chunks in a shard, the index would look like::
+Empty inner chunks are denoted by setting both offset and nbytes to ``2^64 - 1``. 
+Empty inner chunks are interpreted as being filled with the fill value. The index 
+always has the full shape of all possible inner chunks per shard, even if they extend
+beyond the array shape.
+
+The index is placed at the end of the file and encoded into binary representations 
+using the specified index codecs. The byte size of the index is determined by the 
+number of inner chunks in the shard ``n``, i.e. the product of the sizes specified in
```
jbms, line 189: It does not depend on the `chunk_shape`, but rather then `sub_chunk_grid_shape`, equal to the outer chunk shape divided by the inner chunk shape.

Also we should just say that it is determined based on the `sub_chunk_grid_shape` and the combination of index codecs.  While the product of the sizes in `sub_chunk_grid_shape` will almost surely figure into it, that is ultimately up to the codecs.

jbms:
Thanks!  Looks good to me other than the comment I raised.

mkitti:
I'm a bit confused how `c.compute_encoded_size` will work. To use this, I have to compute the byte size of the decoded representation? Is there some function to compute the size of the decoded representation?

Rather I think the input should just be `n`, the number of inner chunks. Perhaps there should also be a `c.compute_decoded_size` as well which also takes `n` as an input.

jbms:
> I'm a bit confused how `c.compute_encoded_size` will work. To use this, I have to compute the byte size of the decoded representation? Is there some function to compute the size of the decoded representation?

I think this could indeed be clarified, since we have "array -> array", "array -> bytes", and "bytes -> bytes" codecs to consider.  For "array -> array" and "array -> bytes", the input is an array and therefore the relevant size would be the shape of the array, not its size in bytes.  For "bytes -> bytes" codecs the input size would be in bytes.
> 
> Rather I think the input should just be `n`, the number of inner chunks. Perhaps there should also be a `c.compute_decoded_size` as well which also takes `n` as an input.

normanrz:
> I'm a bit confused how `c.compute_encoded_size` will work. To use this, I have to compute the byte size of the decoded representation? Is there some function to compute the size of the decoded representation?

The size of the decoded representation is `product(shape) * dtype.itemsize`. The shape would be the chunk shape or index shape depending on the context. `c.compute_encoded_size` can be implemented for all codecs that produce non-variable-length outputs.

Please note, that these are just implementation suggestions. Some implementations may choose to do this differently.

normanrz:
> > I'm a bit confused how `c.compute_encoded_size` will work. To use this, I have to compute the byte size of the decoded representation? Is there some function to compute the size of the decoded representation?
> 
> I think this could indeed be clarified, since we have "array -> array", "array -> bytes", and "bytes -> bytes" codecs to consider. For "array -> array" and "array -> bytes", the input is an array and therefore the relevant size would be the shape of the array, not its size in bytes. For "bytes -> bytes" codecs the input size would be in bytes.

That makes sense. I'll clarify that.

normanrz:
> For "array -> array" and "array -> bytes", the input is an array and therefore the relevant size would be the shape of the array, not its size in bytes.

Actually, this can be complicated for `array -> array` codecs that change the dtype. For these case, `c.compute_encoded_size` would need to track the dtype.

jbms:
Yes, the "array -> array" codec should perhaps also indicate the output dtype.

Possibly it is easier to just leave these implementation details unspecified.

docs/v3/codecs/crc32c/v1.0.rst
```diff
@@ -0,0 +1,95 @@
...
+Configuration parameters
+========================
+
+None.
+
+
+Format and algorithm
+====================
+
+This is an ``bytes -> bytes`` codec.
```
jstriebel, line 72:
```suggestion
This is a ``bytes -> bytes`` codec.
```

jstriebel (dismissed):
LGTM, the generalization of the shard index representation is great :tada:

normanrz:
Thanks for the feedback! I added some more text to clarify the `c.compute_encoded_size` signatures.

jbms:
> The documentation looks improved.
> 
> What is the current method to identify if a codec is one of `array -> array`, `array -> bytes`, or `bytes -> bytes`?

You can't tell just from looking at the json, but any supported codec must be known to the implementation, and it is up to the implementation exactly how this is managed.

In tensorstore there is a class for each codec, which inherits from one of `ZarrArrayToArrayCodecSpec`, `ZarrArrayToBytesCodecSpec`, or `ZarrBytesToBytesCodecSpec`; these in turn all inherit from `ZarrCodecSpec`, which has a virtual method `ZarrCodecKind kind()` where `ZarrCodecKind` is `enum ZarrCodecKind { kArrayToArray, kArrayToBytes, kBytesToBytes };`.

To parse the json codec list, it first parses as a generic list of `ZarrCodecSpec` and then sorts them into a separate list of array -> array codecs, the array -> bytes codec, and a list of bytes -> bytes codecs.

jstriebel (approved):

docs/v3/core/v3.0.rst
```diff
@@ -1186,10 +1186,19 @@ of an additional operation:
...
-- ``c.compute_encoded_size(input_size)``, a procedure that determines the byte 
-  size of the encoded representation given a byte size of the decoded representation.
+- ``c.compute_encoded_size(input_size)``, a procedure that determines the 
+  size of the encoded representation given a size of the decoded representation.
   This procedure cannot be implemented for codecs that produce variable-sized
-  encoded representations, such as compression algorithms.
+  encoded representations, such as compression algorithms. Depending on the 
+  type of the codec, the signature could differ:
+
+  - ``c.compute_encoded_size(array_size, dtype) -> (array_size, dtpye)``
```
jbms, line 1195: typo, "dtpye"

normanrz:
Please review and merge if approved. Thanks!

jbms (approved):

PR #256, opened 2023-08-01
Add zstd codec

jbms:


jbms:
@normanrz Please take a look.

normanrz (approved):

jstriebel (commented):
Should this still be part of ZEP 1? I'd argue for a mini-ZEP for this. IMO it's a good candidate to show that ZEPs can be quite small and concise, with a short duration for acceptance.

Apart from that LGTM!

jbms:
Agreed that this should be separate from ZEP 1.  I should update this document to not mention ZEP 1.  But perhaps there should be a lighter weight process --- it would be nice to just get a vote on this PR directly, rather than writing up a separate ZEP document, etc.

There are going to be a lot of codecs added, and it would be nice to make that process relatively easy while still getting appropriate feedback.

jstriebel:
> But perhaps there should be a lighter weight process --- it would be nice to just get a vote on this PR directly, rather than writing up a separate ZEP document, etc.

That would indeed be nice, but maybe we could have lean ZEPs, with less boilerplate and just pointing to an MR?

@MSanKeys963, what's your take on this?

mkitti (approved):

mkitti:
Is there any chance we could change the default compression level to match `ZSTD_CLEVEL_DEFAULT` which is 3?
https://github.com/facebook/zstd/blob/97291fc5020a8994019ab76cf0cda83a9824374c/lib/zstd.h#L129

jbms:
This proposal doesn't indicate a default compression level if `level` is unspecified --- the `level` parameter is intended to be required, though I should clarify that.

When creating an array, implementations are free to provide a mechanism for the user to leave the level unspecified, however, and in that case the implementation must choose a default value.

The only note on default is that level 0 means the default level --- that follows the zstd C API.

If the array is intended to be used for writing, then I don't see much value in allowing the configuration options to be unspecified; that just leaves room for implementation variance.  However, if someone is creating zarr metadata to somehow match some existing chunked data, that is known to be zstd compressed (but for which the level may not even be known), and which is intended to be read only, then the level is indeed not needed and it is a bit unfortunate that a fake value would have to be specified.  Still I don't think we really need to optimize for this case.

mkitti:
Just to be clear, `numcodecs.zstd.Zstd` does document that the compression level defaults to `1` and sets `DEFAULT_CLEVEL` to `1`. Default to the C header value would be a change from the current practice in Python's numcodecs.

https://numcodecs.readthedocs.io/en/stable/zstd.html#numcodecs.zstd.Zstd
https://github.com/zarr-developers/numcodecs/blob/79a0e933d78a51f32237b29fc375d0ee27aace37/numcodecs/zstd.pyx#L54

mkitti:
Furthermore, Python's numcodecs seems to disallow negative compression levels and interprets any nonpositive compression level as `1`.

https://github.com/zarr-developers/numcodecs/blob/79a0e933d78a51f32237b29fc375d0ee27aace37/numcodecs/zstd.pyx#L83-L84

jbms:
Default values are only really needed to support evolving the spec.  Default values in the initial version of the spec basically just serve to save space in the stored metadata, which I think is pretty negligible in this case.

The zarr-python v3 implementation is still free to choose whichever parameters it likes if the user passes in `zarr.codecs.Zstd()` for example.  Any defaults indicated in the spec would only apply when reading the stored metadata.

Supporting negative levels should be a trivial fix.

mkitti:
I think the fixes are trivial. The question is would it be considered breaking to change the default compression level?

It's not even clear that we need to include the compression level or the checksum flag in the codec configuration since that information is stored by the codec itself. The compression level is perhaps only really useful if someone is trying to add new compressed chunks to the dataset or if they are trying to update chunks. In that case we only need the information to be consistent with the rest of the dataset.

normanrz:
> I think the fixes are trivial. The question is would it be considered breaking to change the default compression level?

I don't think the spec for the zstd codec should have default values. All options should be mandatory to set. As @jbms outlined, implementations can choose default values. I'll change the [default level to 3 for zarr-python](https://github.com/zarr-developers/zarr-python/pull/1838/commits/da8866409457fbf38ab4aa35a3354c54d83331ef).

> It's not even clear that we need to include the compression level or the checksum flag in the codec configuration since that information is stored by the codec itself. The compression level is perhaps only really useful if someone is trying to add new compressed chunks to the dataset or if they are trying to update chunks. In that case we only need the information to be consistent with the rest of the dataset.

The `gzip` and `blosc` codec also have mandatory level options (among others) for the reason you mentioned. I think the `zstd` codec should be consistent with that.

jbms:
We could permit the parameters to be left unspecified in the metadata if only read operations are performed, and fail if any write operations are performed.  However it is not clear that the added complexity of that is worth it --- it really just saves having to put in some arbitrary values if you don't know the level and only care about reading.

zoj613:
Is the `checksum` parameter really necessary here? I went though the list of implementations in different languages and it seems the large majority do not support adding a checksum to the compressed output. Wouldn't this limit how many Zarr implementations can support this codec's spec?

normanrz:
I know of zstd libraries for Python, C/C++, Java, Javascript and Rust that support the checksum flag. Which languages are you referring to?

zoj613:
> I know of zstd libraries for Python, C/C++, Java, Javascript and Rust that support the checksum flag. Which languages are you referring to?

One example is OCaml. Both libraries available don't support this flag. Looking at the list of implementations at the Zstd official site, the other languages that don't include PHP, Fortran, Swift, Ruby, R, Perl, Common Lisp, Ada, Haskell, Julia, Racket, Nim, Elixir.

mkitti:
In Julia, it's available through the low-level interface:
https://github.com/JuliaIO/CodecZstd.jl/blob/master/src%2FLibZstd_clang.jl#L154

LDeakin:
`zstd` has been chosen as the default codec in `zarr-python` v3.0.0. I find that a little concerning since this has not been merged. What needs to be done to get this PR over the line? Implementations that don't support checksum can just reject arrays with `checksum: true` until they can support it.

Related: https://github.com/zarr-developers/zarr-python/issues/2647

docs/v3/codecs/zstd/v1.0.rst
```diff
@@ -0,0 +1,109 @@
...
+==========
+
+The value of the ``name`` member in the codec object MUST be ``zstd``.
+
+
+Configuration parameters
+========================
+
+level:
+    An integer from -131072 to 22 which controls the speed and level
```
nhz2, line 65: I don't think these levels should be hard coded as they may change in future libzstd versions. libzstd will clamp out-of-range compression level values to the [range it supports](https://github.com/facebook/zstd/blob/a610550e2c05cd08842e173bbeb830f596fdfaeb/lib/compress/zstd_compress.c#L639-L649), so any `int` value of level should be accepted to improve forward compatibility.

docs/v3/codecs/zstd/v1.0.rst
```diff
@@ -0,0 +1,109 @@
...
+    of compression (has no impact on decoding).  A value of 0 indicates to use
+    the default compression level.  Otherwise, a higher level is expected to
+    achieve a higher compression ratio at the cost of lower speed.
+
+checksum:
+    A boolean that indicates whether to store a checksum when writing that will
```
rouault, line 71: Trying to implement that from scratch, it wasn't immediately obvious to me what that checksum was, whether this was some CRC manually appended or something really belonging to libzstd.
Seeing https://github.com/zarr-developers/numcodecs/blob/main/numcodecs/zstd.pyx#L117, it is the later. It could make sense to be explicit about that

mkitti:
I think we need to add some clarity here regarding the pledged size.

My recommendation is that all implementations MUST include the pledged size the Zstandard header. However, they SHOULD be able to decompress data which does not include the pledged size in the Zstandard header.

xref: https://github.com/zarr-developers/numcodecs/pull/707

mkitti:
Should a Zstandard dictionary also be considered here?

nhz2:
Zstandard dictionaries add enough complexity for it to be a new codec. Adding dictionary support is also not backwards compatible when decoding, as a decoder that doesn't know how to read the dictionary will fail to decode the data.

RFC8878 already does a good job at describing how a compliant compressor and a compliant decompressor should act, so I don't think we need to add extra requirements about pledged size or other parameters here. It is fairly easy to decode any valid ZSTD-compressed data, but if an implementation wants to limit itself to only decode a subset of ZSTD, as long as there are clear error messages, that is fine.

mkitti:
[RFC8878 leaves a lot of room for compliance](https://datatracker.ietf.org/doc/html/rfc8878#name-compression-algorithm):

> Unless otherwise indicated below, a compliant compressor must produce data sets that conform to the specifications presented here. However, it does not need to support all options.

> A compliant decompressor must be able to decompress at least one working set of parameters that conforms to the specifications presented here. It may also ignore informative fields, such as the checksum. Whenever it does not support a parameter defined in the compressed stream, it must produce an unambiguous error code and associated error message explaining which parameter is unsupported.

We should be slightly stricter here as I have been involved in addressing the inability to decode unknown frame content sizes across four implementations thus far. Zarr implementations should define which subset of options and parameters should be supported for communication between them. It does not work well if one implementation write zstandard in a way that another implementation cannot decode.

nhz2:
We can have a collection of example zstandard encoded data with different encoding option edge cases for implementations to test against, for example, skippable frames, multiple frames, missing frame content size, and large window sizes. We can then see what the current status is for the various zarr implementations.

mkitti:
Here are the issues that I encountered.

1. Tensorstore encodes Zarr v2 data with an unknown frame content size when using Zstandard:
     - Reported issue: https://github.com/google/tensorstore/issues/182 (unresolved)
2. Neuroglancer could not decode data with an unknown frame content size
     - Reported issue: https://github.com/google/neuroglancer/issues/625 (resolved)
     - See numcodecs.js resolution
3. Numcodecs.js could not decode data with an unknown frame content size:
     - Reported issue: https://github.com/manzt/numcodecs.js/issues/46 ([resolved](https://github.com/manzt/numcodecs.js/issues/46))
4. Python numcodecs could not decode data with an unknown frame content size:
     - Reported issue: https://github.com/zarr-developers/numcodecs/pull/707 (resolved)
     - https://github.com/zarr-developers/numcodecs/issues/424 (resolved)
5. python-zstd does not decode data with an unknown frame content size:
     - Reported issue: https://github.com/sergey-dryabzhinsky/python-zstd/issues/53 (will not resolve)
6. libcramjam did not encode data with a frame content size
     - https://github.com/cramjam/libcramjam/issues/27 (resolved)

jbms:
This is now an extension.

PR #257, opened 2023-08-05
Add bfloat16 data type

jbms:


jstriebel:
I would have assumed that this would be a separate spec under [data-types](https://zarr-specs.readthedocs.io/en/latest/v3/data-types.html), since it's optional.

I can see that this is a lot of boilerplate for a rather simple data type, but on the other hand not all future optional data types should be in the core spec IMO. So I'd draw the line at optional data-types.

MSanKeys963:
Thanks for sending the PR, @jbms and reviewing @jstriebel.

@jbms - do you think `bfloat16` should be under data-types?

jbms:
This is now in zarr-extensions.

PR #258, opened 2023-08-08
V3 is no longer in draft state!

jhamman:
This PR makes editorial changes to the V3 spec page to communicate that the spec has been approved via ZEP0001. 

It also removes the DRAFT watermark. I found, as I assume @jstriebel did, that it is non-trivial to apply the watermark on specific pages.


jstriebel:
Good point, thanks for preparing that @jhamman! I had another look at marking individual pages as a draft in the meantime, and combined this with your changes in #259. Would be great if you could have a look at it.

jhamman:
closed via #260

PR #259, opened 2023-08-10
Mark ZEP 1 as accepted.

jstriebel:
* Remove sphinxmark extension.
* Add custom draft mark mechanism to ZEP 2 pages.
* Update status of specs and consolidate headers.

Built upon #258.


jstriebel:
Deleting this in favor of #260. This is the same content, just created the MR within this repo, to have automatic CI builds for the preview.

PR #263, opened 2023-08-21
Rename endian codec to bytes

normanrz:
Renames the `endian` to `bytes` based on this discussion: https://github.com/scalableminds/zarrita/issues/8#issuecomment-1676251849

### DIFF
--- a/docs/v3/codecs/bytes/v1.0.rst
+++ b/docs/v3/codecs/bytes/v1.0.rst
@@ -62,7 +62,11 @@ Configuration parameters
 ========================
 
 endian:
-    Required for multi-byte data types.  A string equal to either ``"big"`` or ``"little"``.
+    Required for data types for which endianness is applicable. For example, 
+    this includes multi-byte data types, such as ``uint16`` and ``int32``, 
+    but not single-byte data types, such as ``uint8`` or ``bool``. 
+    If present, the value MUST be a string equal to either ``"big"`` or 
+    ``"little"``.
 
 
 Format and algorithm
--- a/docs/v3/codecs/bytes/v1.0.rst
+++ b/docs/v3/codecs/bytes/v1.0.rst
@@ -7,13 +7,13 @@
   **Editor's draft 26 July 2019**
 
 Specification URI:
-    https://zarr-specs.readthedocs.io/en/latest/v3/codecs/endian/v1.0.html
+    https://zarr-specs.readthedocs.io/en/latest/v3/codecs/bytes/v1.0.html
 Corresponding ZEP:
     `ZEP 1 — Zarr specification version 3 <https://zarr.dev/zeps/draft/ZEP0001.html>`_
 Issue tracking:
     `GitHub issues <https://github.com/zarr-developers/zarr-specs/labels/codec>`_
 Suggest an edit for this spec:
-    `GitHub editor <https://github.com/zarr-developers/zarr-specs/blob/main/docs/v3/codecs/endian/v1.0.rst>`_
+    `GitHub editor <https://github.com/zarr-developers/zarr-specs/blob/main/docs/v3/codecs/bytes/v1.0.rst>`_
 
 Copyright 2020 `Zarr core development team
 <https://github.com/orgs/zarr-developers/teams/core-devs>`_. This work
@@ -27,7 +27,8 @@ Abstract
 ========
 
 Defines an ``array -> bytes`` codec that encodes arrays of fixed-size numeric
-data types as little endian or big endian in lexicographical order.
+data types as a sequence of bytes in lexicographical order. For multi-byte data
+types, it encodes the array either in little endian or big endian.
 
 
 Status of this document
@@ -55,14 +56,14 @@ this specification are introduced with the words "for example".
 Codec name
 ==========
 
-The value of the ``name`` member in the codec object MUST be ``endian``.
+The value of the ``name`` member in the codec object MUST be ``bytes``.
 
 
 Configuration parameters
 ========================
 
 endian:
-    Required.  A string equal to either ``"big"`` or ``"little"``.
+    Required for multi-byte data types.  A string equal to either ``"big"`` or ``"little"``.
 
 
 Format and algorithm
@@ -131,4 +132,5 @@ References
 Change log
 ==========
 
-No changes yet.
+- ``endian`` codec was renamed to ``bytes`` codec.  `PR #263
+  <https://github.com/zarr-developers/zarr-specs/pull/263/>`_
--- a/docs/v3/codecs/bytes/v1.0.rst
+++ b/docs/v3/codecs/bytes/v1.0.rst
@@ -1,7 +1,7 @@
-.. _endian-codec-v1:
+.. _bytes-codec-v1:
 
 ============================
- Endian codec (version 1.0)
+ Bytes codec (version 1.0)
 ============================
 
   **Editor's draft 26 July 2019**
--- a/docs/v3/codecs/sharding-indexed/v1.0.rst
+++ b/docs/v3/codecs/sharding-indexed/v1.0.rst
@@ -101,7 +101,7 @@ Sharding can be configured per array in the :ref:`array-metadata` as follows::
             "chunk_shape": [32, 32],
             "codecs": [
               { 
-                "name": "endian",
+                "name": "bytes",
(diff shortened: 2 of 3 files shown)


normanrz:
@jbms @jstriebel please take a look.
Can we set up a redirect from the `endian` page to the new `bytes` page?

rabernat:
:+1: I'm all for this change.

But I'd also like us to understand what is happening here at a procedural level. How is it possible to change the core spec now that it has been accepted? At what point is a new ZEP required?

I agree we should make room for minor changes like this. However, we should do so under a clear framework. How are we tracking different versions of the spec if the spec continues to evolve? Is this a step towards, e.g. v3.0.1?

I have advocated for a more incremental approach to the spec development, as practiced, for example, by [STAC](https://github.com/radiantearth/stac-spec/releases). But our current process doesn't seem to allow for that. We should change that.

rabernat:
> That means there is some room to make some minor changes to the spec that are motivated by feedback from the implementors

That makes sense to me.

STAC released several 1.0.0 "beta" and "rc" releases of the spec before stamping a final 1.0.0. (https://github.com/radiantearth/stac-spec/releases)

Maybe it would be helpful for us to do the same? Otherwise, I worry that we will repeat the current situation where there are many different, formally indistinguishable versions of v3 in the wild.

But we can continue that discussion elsewhere. Sorry for hijacking this thread.

rabernat (approved):

joshmoore:
:+1: for having a clear idea of how this will be communicated to those who have already started implementing since there won't be, e.g., a clear vote point where we get their affirmation.

jbms:
@zarr-developers/implementation-council This is a (minor) breaking change, so approval from existing implementations is needed.

manzt:
**Yes** from me.

mkitti:
While I think `bytes` is a better than `endian`, I wonder if it could be made more obvious that this also uses C-order when converting an array to bytes.

mkitti:
To make this a non-breaking change, could the `endian` codec be deprecated, use is discouraged pending removal, rather than removed?

jbms:
Do you mean a name like bytes_corder?  The idea is that you can combine this with the transpose codec to obtain any order.

mkitti:
I'm not sure I would say "any order", but yes I understand that this is meant to be used with `transpose` to obtain F-order. In the past, "Morton-order" was also discussed, and I'm pretty sure that `transpose` does not allow for that.

My primary concern here is that `bytes` still seems to be less than obvious what what this codec does. 

Looking at this from the ZEP2 perspective, the example is currently as follows.
```json
        "index_codecs": [
          {
            "name": "endian",
            "configuration": {
              "endian": "little",
            }
          },
          { "name": "crc32c"}
        ]
```
I agree that this is confusing. It is unclear that `endian` codec is flattening the array and serializing to bytes.

Now, with this PR, we have the following.
```json
        "index_codecs": [
          {
            "name": "bytes",
            "configuration": {
              "endian": "little",
            }
          },
          { "name": "crc32c"}
        ]
```
While this is better, it is still not quite obvious to someone who has not read the specification thoroughly what this codec operation has performed. If you were a Python user and knew that `bytes(some_numpy_array)` return a `bytes` object, I think there is a good chance one could guess the codec operation. Otherwise, I'm really not sure if this operation could be easily inferred.

If I'm allowed to bikeshed, the names in my head are `ravel_to_bytes` after the numpy function [`ravel`](https://numpy.org/doc/stable/reference/generated/numpy.ravel.html#numpy.ravel) or `flatten_to_bytes`. I think these names capture that some kind of ordering operation is occuring as well as the output.

> At this point, tensorstore, zarrita, and zarrita.js have already made this change. The only other implementation of which I'm aware, zarr3-rs, has not, but I don't know if it is in production use yet.

I'm guessing that those who have made this change also still support the "endian" codec? Returning to Ryan's process question, pre-implementation makes this seem like it has been decided already and perhaps already hard to change.

manzt:
> I'm guessing that those who have made this change also still support the "endian" codec? 

I switched over to “bytes” only in zarrita.js to align with zarrita. I don’t have much opinion here, but happy to implement/support what is decided.

jbms:
I think flatten_to_bytes is not unreasonable, but also is essentially a description of what any array to bytes codec must do, and therefore may not be particularly helpful to identify this specific codec.

mkitti:
It is not clear to me that every array to bytes transformation must be reversible or that lossy compression is prohibited at this stage. One could define an array to bytes transformation that retains no information at all or perhaps only the shape of the array. Decoding would just produce am array of zeros.

While I agree that `flatten_to_bytes` is a trivial loseless transformation, I'm unclear if that truly describes the entire array to bytes codec space.

normanrz:
I don't see much added value from `flatten_to_bytes` over just `bytes`. I think `bytes` captures pretty well what the codec does. If it it not clear from the name, the specification contains all the details.
Considering that some implementations have already adopted it (I want to added Webknossos to Jeremy's list), if we were to change the name now, I think the specification should mandate implementations to handle the alias. Changing from `endian` to `bytes` already has caused some confusion.

> In the past, "Morton-order" was also discussed, and I'm pretty sure that `transpose` does not allow for that.

`transpose` doesn't allow for that, but there could be a `morton` codec that flattens the array in morton order and leaves the byte serialization to `bytes`.

mkitti:
```
        "index_codecs": [
          {
            "name": "transpose",
            "configuration": {
              "order": "C",
             }
          },
          {
            "name": "bytes",
            "configuration": {
              "endian": "little",
            }
          },
          { "name": "crc32c"}
        ]
```

```
        "index_codecs": [
          {
            "name": "transpose",
            "configuration": {
              "order": "F",
             }
          },
          {
            "name": "bytes",
            "configuration": {
              "endian": "little",
            }
          },
          { "name": "crc32c"}
        ]
```

jbms:
Yes, that's right, though of course the other PR about how `order` is specified impacts that.

Note, for example, that if there is a transpose codec specified prior to the sharding codec, the index, the chunk_shape, and the data chunks are then relative to the transposed order.

DennisHeimbigner:
May I suggest that the configuration.endian key also allow "native" in addition to "big" and "little".

DennisHeimbigner:
Shouldn't the configuration for endian also specify the dtype so that it is self-contained?

jbms:
Note that this is specifying the stored (i.e. "on-disk") format --- for that purpose, "native" is problematic because the native endianness of the machine reading it may differ from the native endianness of the machine writing (though of course in practice it will almost always be little endian).  When creating an array, though, a zarr implementation may provide a way to choose native endianness automatically.  E.g. in tensorstore, you can leave "endian" unspecified and it will choose the native endianness (but when actually stored, the "endian" value will always be specified).

As far as storing the dtype, I can see how that may be advantageous though a similar argument would apply to storing the shape of the chunk as well in the `bytes` codec configuration, and it is not clear that we want to duplicate all of that.

DennisHeimbigner:
Fair enough.

jstriebel (dismissed):

clbarnes:
Ah yes, this slipped past me. The name change isn't a concern; making the order optional opens up some new codec errors types conditional on dtype, which is a bit of a pain (damn rust making you think about these things early...) but shouldn't be too bad.

clbarnes:
Do multi-byte raw codecs have an endianness, i.e. do they require the config here?

clbarnes:
Does that get complicated where they're used as a fallback? Out of scope for this PR, I suppose.

Endian codec rename PR ready to go, anyway https://github.com/clbarnes/zarr3-rs/pull/18

jbms:
> Does that get complicated where they're used as a fallback? Out of scope for this PR, I suppose.

Yes it is a complication, and indeed it is for reasons like this that we removed the concept of fallback data types.

> 
> Endian codec rename PR ready to go, anyway [clbarnes/zarr3-rs#18](https://github.com/clbarnes/zarr3-rs/pull/18)

jbms:
Now that this has been implemented by all existing zarr v3 implementations, I think it is time to merge this.

jbms:
@normanrz I think this should be merged --- can you resolve the conflicts so that we can merge it?

normanrz:
> can you resolve the conflicts so that we can merge it?

Done!

jbms (approved):

jbms (approved):

PR #264, opened 2023-09-11
Require transpose codec `order` option to be specified as an explicit permutation

jbms:
The "C" and "F" constants were added to emphasize the relationship with the zarr v2 `order` metadata field.  However, these contants are somewhat confusing:

- "C" means no-op and therefore never needs to be used

- "F" means reverse dimension order.  If there are no other changes to the dimesnion order, this does indeed result in Fortran-order storage, but if there are other changes to the dimension order, that is not the case.  For example, in the following:

```json
"codecs": [
  {"name": "transpose", {"configuration": {"order": "F"}}},
  {"name": "transpose", {"configuration": {"order": "C"}}},
  {"name": "bytes"}
]
```

or

```json
"codecs": [
  {"name": "transpose", {"configuration": {"order": "C"}}},
  {"name": "transpose", {"configuration": {"order": "F"}}},
  {"name": "bytes"}
]
```

the storage order is actually "F", while in the following:

```json
"codecs": [
  {"name": "transpose", {"configuration": {"order": "F"}}},
  {"name": "transpose", {"configuration": {"order": "F"}}},
  {"name": "bytes"}
]
```

the storage order is actually "C".

### DIFF
--- a/docs/v3/codecs/transpose/v1.0.rst
+++ b/docs/v3/codecs/transpose/v1.0.rst
@@ -65,15 +65,9 @@ Configuration parameters
 ========================
 
 order:
-    Required.  Must be one of:
-
-    - An array of integers specifying a permutation of ``0``, ``1``, ...,
-      ``n-1``, where ``n`` is the number of dimensions in the decoded chunk
-      representation provided as input to this codec.
-    - The string ``"C"``, equivalent to specifying the identity permutation
-      ``0``, ``1``, ..., ``n-1``.  This makes the codec a no-op.
-    - The string ``"F"``, equivalent to specifying the permutation ``n-1``, ...,
-      ``1``, ``0``.
+    Required.  Must be an array of integers specifying a permutation of ``0``, ``1``, ...,
+    `n-1``, where ``n`` is the number of dimensions in the decoded chunk
+    representation provided as input to this codec.
 
 Format and algorithm
 ====================
@@ -105,4 +99,8 @@ References
 Change log
 ==========
 
-No changes yet.
+Changes after acceptance of ZEP 1
+---------------------------------
+
+The ``order`` configuration parameter no longer supports the constants ``"C"``
+or ``"F"`` and must instead always be specified as an explicit permutation.


mkitti:
To me this sounds like a new codec called `permutedims` which is reminiscent of a [julia function](https://docs.julialang.org/en/v1/base/arrays/#Base.permutedims) of the same name. There's also a MATLAB function [`permute`](https://www.mathworks.com/help/matlab/ref/permute.html?s_tid=doc_ta#f93-999230) and a `R` function [`aperm`](https://rdrr.io/r/base/aperm.html).

From your example, it does seem like that use of multiple `transpose` codecs in a row should not be permitted and perhaps that it should only be applied first.

jbms:
> To me this sounds like a new codec called `permutedims` which is reminiscent of a [julia function](https://docs.julialang.org/en/v1/base/arrays/#Base.permutedims) of the same name. There's also a MATLAB function [`permute`](https://www.mathworks.com/help/matlab/ref/permute.html?s_tid=doc_ta#f93-999230) and a `R` function [`aperm`](https://rdrr.io/r/base/aperm.html).
> 
> From your example, it does seem like that use of multiple `transpose` codecs in a row should not be permitted and perhaps that it should only be applied first.

To be clear, the existing spec already allows an explicit permutation, I am just proposing to not allow "C" and "F" as aliases.

I would agree that permutedims would be a fine name, but happened to choose transpose to match the numpy function.

While applying the transpose codec multiple times consecutively could always be equivalently represented as a single application of transpose with the composed permutation, and implementations might indeed wish to perform that optimization, I don't think it would be helpful to actually disallow specifying transpose multiple times; that might reasonably occur if an automated tool modifies the metadata and doesn't handle composing the transposes itself.   In any case even if multiple consecutive transposes are disallowed the issue can still occur if there are other codecs in between, such as delta encoding, sharding, etc.

normanrz:
After giving it some thought, I agree to this change. In fact, I also had some headaches to make `F` and `C` aliases work in zarrita. Allowing just the array notation would make it clearer. I'm fine with sticking with `transpose` as a name for the codec.

jstriebel (approved):
LGTM :+1: No strong opinion about the name, `transpose` seems fine to me.

We might want to add a section somewhere explaining different options when dealing with differently ordered arrays (either on disk (or in the cloud) or in memory). I think this covers all possible options:
* When disk representation is different from the expected in-memory representation, one might consider to present the dimensions in a different order to the user than they are stored on-disk.
  *Example*: The array is stored on-disk in F-order, and the in-memory representation must be in C order. Then the array might be loaded in F order, but presented to the user in the reversed order. In this case the order of the `dimension_names` field would need to be reversed, and this functionality must be available across the whole implementation.
* If the in-memory representation is flexible withouth changing the semantic order presented to the user (this is the case e.g. for numpy), one might choose to use the on-disk order in this case.
  *Example*: The array is stored on-disk in F-order, and the in-memory representation is a numpy array, without specifying that C order is expected. Then the array might be loaded in F order. The user still sees the array in the same order as stored (in the order of the `dimension_names` field).
* The last option is to change the data-layout and adapt the exptected in-memory layout.

docs/v3/codecs/transpose/v1.0.rst
```diff
@@ -105,4 +99,8 @@ References
 Change log
 ==========
 
-No changes yet.
+Changes after acceptance of ZEP 1
+---------------------------------
+
+The ``order`` configuration parameter no longer supports the constants ``"C"``
```
joshmoore, line 105: Along with other late changes, @jbms, we probably should send out a heads up of some form for all implementers.

cc: @MSanKeys963

PR #271, opened 2023-10-29
spec: add a motivation for the c prefix on chunk keys

d-v-b:
I added a motivation for the `c` prefix on chunk keys. I'm happy to modify this if the reported motivation is factually or stylistically incorrect :)

### DIFF
--- a/docs/v3/core/v3.0.rst
+++ b/docs/v3/core/v3.0.rst
@@ -539,7 +539,7 @@ mandatory names:
          in which case the coordinate of a chunk is the empty tuple, and the chunk key
          will consist of the string ``c``.
 
-      .. note:: In the ``default`` ```chunk_key_encoding`` described above, 
+      .. note:: In the ``default`` ``chunk_key_encoding`` described above, 
          chunk keys are stored under the prefix ``c```, which separates chunk keys from 
          the keys for metadata documents. This is a performance optimization that 
          allows applications to find Zarr metadata documents without traversing a 
--- a/docs/v3/core/v3.0.rst
+++ b/docs/v3/core/v3.0.rst
@@ -529,7 +529,7 @@ mandatory names:
 
       For example, in a 3 dimensional array, with a separator of ``/`` the identifier
       for the chunk at grid index (1, 23, 45) is the string ``"c/1/23/45"``.  With a
-      separator of ``.``, the identifier is the string ``"c.1.23.45"``. The inital prefix 
+      separator of ``.``, the identifier is the string ``"c.1.23.45"``. The initial prefix 
       ``c`` ensures that metadata documents and chunks have separate prefixes.
 
       .. note:: A main difference with spec v2 is that the default chunk separator
--- a/docs/v3/core/v3.0.rst
+++ b/docs/v3/core/v3.0.rst
@@ -529,7 +529,8 @@ mandatory names:
 
       For example, in a 3 dimensional array, with a separator of ``/`` the identifier
       for the chunk at grid index (1, 23, 45) is the string ``"c/1/23/45"``.  With a
-      separator of ``.``, the identifier is the string ``"c.1.23.45"``.
+      separator of ``.``, the identifier is the string ``"c.1.23.45"``. The inital prefix 
+      ``c`` ensures that metadata documents and chunks have separate prefixes.
 
       .. note:: A main difference with spec v2 is that the default chunk separator
          changed from ``.`` to ``/``, as in N5.  This decreases the maximum number of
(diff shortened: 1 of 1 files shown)


docs/v3/core/v3.0.rst
```diff
@@ -539,6 +539,12 @@ mandatory names:
          in which case the coordinate of a chunk is the empty tuple, and the chunk key
          will consist of the string ``c``.
 
+      .. note:: In the ``default`` ```chunk_key_encoding`` described above,
```
jstriebel, line 542:
```suggestion
      .. note:: In the ``default`` ``chunk_key_encoding`` described above,
```

jstriebel (dismissed):
LGTM, thank you! :+1:

docs/v3/core/v3.0.rst
```diff
@@ -539,6 +539,12 @@ mandatory names:
          in which case the coordinate of a chunk is the empty tuple, and the chunk key
          will consist of the string ``c``.
 
+      .. note:: In the ``default`` ``chunk_key_encoding`` described above, 
+         chunk keys are stored under the prefix ``c```, which separates chunk keys from
```
LDeakin, line 543: there is an extra \` after the c

d-v-b: fixed! thanks for spotting it

LDeakin:
Just checking, doesn't your note only apply when the separator is `/`?

d-v-b:
@LDeakin you're right, the spec indicates that for the dimension separator ".", we should have chunk keys like [`c.1.23.45`](https://github.com/zarr-developers/zarr-specs/blob/97453eb709653124bc511c54ecc5d947f936e93c/docs/v3/core/v3.0.rst?plain=1#L532), which offers no benefit when listing the contents of a directory (and the "." separator will lead to the most objects under one prefix, too). I think either I have misunderstood the purpose of the "c" prefix, or the spec should be altered to be true to the purpose indicated in my note, i.e. by stipulating that chunk keys always start with "c/" instead of "c<dimension_separator>". 

Can you clarify @jbms ?

jbms:
If "c/" were used as the prefix in all cases, then on storage systems that have a concept of directories, to create a zarr array would require a minimum of 2 directories, rather than 1 as in zarr v2, which may be significant added cost when creating a large number of arrays on some storage systems with high per-directory cost.

If you are using "/" as the separator then you will already have to create a lot of directories and one more would probably not be significant.

d-v-b:
@jbms how would you describe the overall purpose of the "c" prefix then?

jbms:
I wasn't strongly for or against it.  But there was a desire to have a separate prefix for chunks to keep the key space more organized, and easier for users to browse when using directories.  That is accomplished when using a separator of "/".  When using a separator of "." there is still a separate prefix but for directory-based stores it won't be too helpful, but that way we have the option to avoid the overhead of an additional directory, without introducing an additional option.  Using a prefix of "c" also means we don't need a special case for zero dimensions, the single chunk is just "c".  In zarr v2 we use "0" as a special case.

d-v-b:
I have to confess I find this confusing: when the dimension_separator is "/" (i.e., when chunks are stored in separate directories, so the top-level directory contains no chunks at all), the prefix of "c" is used to facilitate browsing directories, but when the dimension_separator is ".", i.e., when all the chunks are in the top-level directory, the prefix of "c" has a different purpose? It seems like the "." case is exactly when directory browsing is most inconvenient. Am I missing something here?

jbms:
Even with "/" you can still have a lot of entries under "c/" if the first dimension has a lot of chunks.  In any case the "c" prefix still serves to separate the key space, even if it doesn't result in a separate directory (and on storage systems like s3 or gcs there are no directories anyway).  In the future, we may likely want to add new key encodings to make certain queries on chunks more efficient, and those key encodings may use characters other than 0-9.  Having a separate prefix of "c" helps avoid conflicts with other extension metadata keys.

d-v-b:
that makes sense; I will edit this PR to just state that the purpose of the "c" is to partition the key space

docs/v3/core/v3.0.rst
```diff
@@ -529,7 +529,8 @@ mandatory names:
 
       For example, in a 3 dimensional array, with a separator of ``/`` the identifier
       for the chunk at grid index (1, 23, 45) is the string ``"c/1/23/45"``.  With a
-      separator of ``.``, the identifier is the string ``"c.1.23.45"``.
+      separator of ``.``, the identifier is the string ``"c.1.23.45"``. The inital prefix
```
joshmoore, line 532:
```suggestion
      separator of ``.``, the identifier is the string ``"c.1.23.45"``. The initial prefix 
```

jbms (approved):

PR #272, opened 2023-10-29
style tweaks

d-v-b:
- changed instances of "zarr" in prose to "Zarr"
- changed instances of "behavior" to "behaviour" (happy to go the other way, too -- we should just be consistent)
- made some prose clearer

### DIFF
--- a/docs/v3/core/v3.0.rst
+++ b/docs/v3/core/v3.0.rst
@@ -839,7 +839,7 @@ apply to node names:
 * must not be a string composed only of period characters, e.g. ``"."`` or ``".."``
 * must not start with the reserved prefix ``"__"``
 
-To ensure consistent behaviour across different storage systems and programming
+To ensure consistent behavior across different storage systems and programming
 languages, we recommend users to only use characters in the sets ``a-z``,
 ``A-Z``, ``0-9``, ``-``, ``_``, ``.``.
 
@@ -1223,7 +1223,7 @@ codec specification must declare the codec identifier, and describe
 (or cite documents that describe) the encoding and decoding algorithms
 and the format of the encoded data.
 
-A codec may have configuration parameters which modify the behaviour
+A codec may have configuration parameters which modify the behavior
 of the codec in some way. For example, a compression codec may have a
 compression level parameter, which is an integer that affects the
 resulting compression ratio of the data. Configuration parameters must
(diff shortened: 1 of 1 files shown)


jstriebel (commented):
Thanks a lot for improving the consistency! LGTM, but I'd rather let a native speaker do the approval ;-) cc @jbms @joshmoore @rabernat

docs/v3/core/v3.0.rst
```diff
@@ -839,7 +839,7 @@ apply to node names:
 * must not be a string composed only of period characters, e.g. ``"."`` or ``".."``
 * must not start with the reserved prefix ``"__"``
 
-To ensure consistent behaviour across different storage systems and programming
+To ensure consistent behavior across different storage systems and programming
```
jbms, line 842: This is a UK vs US English difference, not sure it is worth worrying about.

docs/v3/core/v3.0.rst
```diff
@@ -1649,23 +1649,23 @@ Implementation Notes
 This section is non-normative and presents notes from implementers about cases
 that need to be carefully considered but do not strictly fall into the spec.
 
-Explicit vs implicit group
+Explicit vs implicit groups
```
jbms, line 1652: Need to adjust underlines below

docs/v3/core/v3.0.rst
```diff
@@ -1350,11 +1350,11 @@ A **writeable store** supports the following operations:
 
 .. note::
 
-   Some KV stores do allow creation and update of keys, but not deletion. For
+   Some stores allow creatiing and updating keys, but not deleting them. For
```
jbms, line 1353: typo on creatiing

docs/v3/core/v3.0.rst
joshmoore: Heh. I tend to agree that it's not critical, but I have experienced in other projects that once the topic comes up it *keeps* coming up. We should probably just pick one and put it in the [dev guide](https://github.com/zarr-developers/governance/blob/main/CORE_DEV_GUIDE.md). (And if we can't agree then there cane be a ZEP :wink:)

docs/v3/core/v3.0.rst
joshmoore, line 1353:
```suggestion
   Some stores allow creating and updating keys, but not deleting them. For
```

docs/v3/core/v3.0.rst
d-v-b: yep, for this document we should just pick one style and stick with it. I'm happy do do british english as well. the point is just to be consistent.

docs/v3/core/v3.0.rst
d-v-b: good catch, this should be fixed

docs/v3/core/v3.0.rst
joshmoore: Then since it's 4 to 1 in this file, @d-v-b, let's go with `behaviour` for the moment. (In zarr-python, it's 23 to 19). We will need to decide, but at least we can err on the side of fewer changes for the moment.

Issue #291, opened 2024-04-12
Should we keep implicit groups in Zarr V3?


d-v-b:
The v3 spec permits the existence of Zarr groups without any distinguishing metadata. 

In the [section comparing v3 with v2](https://zarr-specs.readthedocs.io/en/latest/v3/core/v3.0.html#comparison-with-zarr-v2), the spec states 

> v3 allows for greater flexibility in how groups and arrays are created. In particular, v3 supports implicit groups, which are groups that do not have a metadata document but whose existence is implied by descendant nodes. This change enables multiple arrays to be created in parallel without generating race conditions for the metadata when creating parent groups.

So the argument here is that we want to avoid race conditions when creating arrays in parallel. Is this a serious problem for anyone? Personally, I was not aware that parallel hierarchy mutation was a design goal of Zarr. I always thought that the only parallelism guarantees were for separate array chunks; since creating nodes in the hierarchy is so simple (just write a JSON document), there shouldn't be a motivation for parallelizing this process, at least that's how it seems to me. 

Later, there is a section [comparing explicit and implicit groups](https://zarr-specs.readthedocs.io/en/latest/v3/core/v3.0.html#explicit-vs-implicit-groups), which states

> This specification defines both implicit and explicit groups, but implementations may create an explicit group for all implicit groups they encounter, in particular when using a hierarchical storage.
> 
> Erasure of an implicit group may automatically erase any empty parent. For example on a S3 store where the namespace is flat, erasure of the last key with a prefix will erase all implicit groups in the prefix.
> 
> Care must be taken when erasing an array or a group if the parent needs to be converted into an explicit group.
>
>A race-condition arises if a client writes an array at path P, and another client concurrently assumes P is an implicit group and writes subgroups or arrays into it. Implementations can avoid this race condition by exclusively using explicit groups.

So here we learn that implicit groups actually introduce a new type of race condition, because they make the structure of Zarr hierarchy ambiguous, and there's a suggestion that implementations modify Zarr hierarchies they encounter to insert implicit groups when they are detected. I don't think this is great. First, we have traded the race condition that motivated implicit groups for another one, so we are net 0 race conditions, and we are encouraging implementations to mutate the hierarchies they encounter, perhaps as an admission that implicit groups might be a bit of a headache in practice.  

I'm honestly not sure what the advantage is of implicit groups. Here are some disadvantages, from my POV:
[...]

Issue #184, opened 2022-11-28
Implicit vs explicit groups


jbms:
Currently the zarr v3 spec does not require that every group have a metadata document, and instead allows a group to be defined implicitly by the presence of a descendant.

Pros:
- No need to write an empty metadata document just to have a group.
- Multiple machines can create distinct arrays in parallel within the same, possibly not-yet-existing group without any coordination or store-level locking/atomic operations to create the group metadata document.

Cons:
- Possibility of simultaneously creating a group and array with the same name.  This could be fixed by allowing arrays to also be groups.
- If group-level storage transformers are allowed, there is a possible race condition: one machine may attempt to create "a/b" as an array, and another machine may simultaneously create "a" as a group with a storage transformer.  We may end up with a corrupted result, where "a/b" does not correctly take into account the storage transformer.

An alternative to implicit groups is to use a separator other than "/", such as ".", if the equivalent of implicit groups are desired.

jbms:
Also, even if racy due to a lack of support for atomic read/modify/write, writing the empty group files from multiple machines concurrently could still be fine; since the contents would be the same from all writers, it doesn't matter which writer wins.

jstriebel:
> * Possibility of simultaneously creating a group and array with the same name.  This could be fixed by allowing arrays to also be groups.

Since the names of the metadata documents will be both `zarr.json` for arrays and groups as of #200 this case seems to be eliminated.

> * If group-level storage transformers are allowed, there is a possible race condition: one machine may attempt to create "a/b" as an array, and another machine may simultaneously create "a" as a group with a storage transformer.  We may end up with a corrupted result, where "a/b" does not correctly take into account the storage transformer.

True, but I don't see how this can be circumvented. Even if implicit groups use a different delimiter than `/`, both could be created simultaneously, and one of both would win, the other would probably not exist. I don't see a way around this race condition.

@jbms Do you think there is something to do as part of the v3 core spec? Can we close this?

jbms:
Yes we can close this.

PR #292, opened 2024-04-12
remove implicit groups

d-v-b:
this PR removes support for implicit groups (Zarr groups with no metadata documentation) from the spec. See #291 for the motivation. Besides changing the content of the text, I don't know what else to change, so I'm keeping this a draft for now.

A mergeable version of this PR would include a section explaining what implicit groups are, why they were added previously, and why they are now removed. I am happy to add this if it looks like this PR has legs.

Related discussion: #184


MSanKeys963:
@d-v-b, is this PR ready for review?

d-v-b:
@MSanKeys963 yes!

d-v-b:
how can we move forward with this? I think it's important to make these changes quickly, before we get skew between implementations.

jhamman:
@zarr-developers/steering-council - this PR has sat without review for 4 months. Can we move it forward please?

jbms (dismissed):

rabernat (approved):
Thanks for the push @jhamman.

joshmoore:
@d-v-b: do you want to take a look at the conflict?

d-v-b:
> @d-v-b: do you want to take a look at the conflict?

done!

bjoernthiel:
@joshmoore Please add the following line directly after "Group metadata": 
Each Zarr group in a hierarchy must have a group metadata document, named zarr.json.

PR #301, opened 2024-07-05
Fix error in the v3 spec regarding discovering all nodes using `list_prefix`.

zoj613:
The document mentions using `list_prefix("/")` to discover all nodes in a store. But the spec says that a root node has the empty string prefix. So `list_prefix("/")` would return the empty list since no other non-root node can have have the "/" prefix since all non-root nodes need to have a name. This commit fixes that error by replacing the call to `list_prefix` with the empty string.

### DIFF
--- a/docs/v3/core/v3.0.rst
+++ b/docs/v3/core/v3.0.rst
@@ -1549,7 +1549,7 @@ Let "+" be the string concatenation operator.
     groups.
 
     For hierarchies without group storage transformers one may also call
-    ``list_prefix("/")``. All ``zarr.json`` keys represent either explicit
+    ``list_prefix("")``. All ``zarr.json`` keys represent either explicit
     groups or arrays. All intermediate prefixes ending in a ``/`` are implicit
     groups.
 


zoj613:
I also noted that the example JSON array metadata document in the **Array metadata** section is ill-formed. the `codecs` field contains just the gzip codec configuration, even though the spec requires the codec chain to always have one `bytes->bytes` codec. That is in the example the gzip codec should be replaced with `{"name": "bytes", "configuration": "little"}` to be considered correct. Not sure if that's worth a pull request or not. There are other small errors I picked up while reading the spec that I can't immediately recall. Let me know if opening a PR for every error I note is worthwhile.

d-v-b:
> Let me know if opening a PR for every error I note is worthwhile.

Absolutely! Please do if you have the time

zoj613:
> > Let me know if opening a PR for every error I note is worthwhile.
> 
> Absolutely! Please do if you have the time

done, see: https://github.com/zarr-developers/zarr-specs/pull/302

d-v-b (approved):

d-v-b (approved):

joshmoore (approved):

Issue #298, opened 2024-05-31
v3 codec structure in zarr.json


d-v-b:
the v3 spec states that codecs are stored in a JSON array under the key `codecs`. But the spec also states that the list of codecs is [structured](https://github.com/zarr-developers/zarr-specs/blob/daef2bc91a246d24d895276679b6ac9556d7ebb1/docs/v3/core/v3.0.rst?plain=1#L1095-L1100): 
> ...the list of codecs must be of the following form:

    zero or more array -> array codecs; followed by
    exactly one array -> bytes codec; followed by
    zero or more bytes -> bytes codecs.

This is actually a _lot_ of semantic load for something simple like a JSON array. Instead of using a JSON array, I believe that the above structure could be expressed much better (where "expressed better" means "conveys intent more clearly, with no loss of information, and minimal added complexity") by using a JSON object with the following structure:

```json
{
"codecs": {
    "array_array": [], # array of array -> array codecs, possibly empty
    "array_bytes": {"name": "bytes"}, # single of array -> bytes codec, required
    "bytes_bytes": [], # array of bytes -> bytes codecs, possibly empty
}
```

I am noting this because over in the `zarr-python` v3 implementation effort, we have written something like the above data structure as part of the basic parsing of the contents of `zarr.json`. In fact I think this data structure will arise in _any_ implementation, because implementations must represent the structure of the codecs, and that structure is not captured at all by the JSON array representation. But, as I show here, it is trivial to describe the codec structure explicitly with JSON. A corollary benefit is that the above proposed data structure expresses much better the constraint that there be just 1 array -> bytes codec, which would reduce some validation burden from implementations.

So, if we care about making this easier for implementations (and I think making it easy for implementations also makes it easier for users), we should considering this change to `zarr.json`. There is no change to the semantics of the spec, but it makes `zarr.json` more clear. I understand that people may not want to change the spec. But I consider that a separate question from whether the current spec has defects that could in principle be fixed, such as the one described here.

normanrz:
My take is that this kind of change is hard to get in at this point. 

[...]

PR #302, opened 2024-07-06
Fix Example JSON document for Array metadata.

zoj613:
JSON array metadata document in the **Array metadata** section is ill-formed. the `codecs` field contains just the gzip codec configuration, even though the spec requires the codec chain to always have one `array->bytes` codec. That is in the example the gzip codec should be replaced with something more sensible like
`{"name": "bytes", "configuration": "big"}` to be considered correct. This commit fixes that.

### DIFF
--- a/docs/v3/core/v3.0.rst
+++ b/docs/v3/core/v3.0.rst
@@ -721,9 +721,9 @@ compressed using gzip compression prior to storage::
             }
         },
         "codecs": [{
-            "name": "gzip",
+            "name": "bytes",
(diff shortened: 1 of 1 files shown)


d-v-b:
thanks for the fix! I will use this opportunity to suggest that the kind of mistake you fixed here would be less likely if we used a more structured JSON representation for the codecs, c.f. https://github.com/zarr-developers/zarr-specs/issues/298

joshmoore (approved):

PR #304, opened 2024-07-26
Clarify relationship between the inner chunk shape and shard shape

sbesson:
This commit reviews the definitions to enforce that the inner chunk shape must have the same number of dimensions (rather than the same dimensions) as the shard shape with the additional relationship specified by the next sentence.
This change is consistent with the terminology used for the chunk_shape key in the configuration parameters section above


d-v-b (approved):
thanks for the fix!

PR #306, opened 2024-08-11
Fix data storage key exmaple

dstansby:
Near the top of the spec it says:

> All chunk or other data of an array is stored under the key prefix determined by its path. For a root array, the key prefix is obtained from the metadata document key by stripping the trailing zarr.json. For example, for a root array, the prefix is the empty string. For a non-root array with hierarchy path /foo/bar, the prefix is foo/bar/.

As such, I don't think there should be a `c` level in the example I've edited?


dstansby:
Ah maybe not, now I've read the rest of the spec! It's not super clear that in the default implementation that `c` is a special path.

Issue #266, opened 2023-09-26
Byte ranges need to permit reading from the end


LDeakin:
The specification mentions that the `decoded regions` parameter in `c.partial_decode()` specifies a list of byte ranges. With the `sharding` codec, to read the shard index without reading the whole file it must be possible to read some number of bytes from the end of the `input_handle`.

In the [Abstract Store Interface](https://zarr-specs.readthedocs.io/en/latest/v3/core/v3.0.html#abstract-store-interface) section, a byte range is specified as
`To read and write partial values, a range specifies two integers range_start and range_length, that specify a part of the value starting at byte range_start (inclusive) and having a length of range_length bytes. range_length may be none, indicating all available data until the end of the referenced value.`

I think that paragraph should be updated to reflect that a byte range can be
- relative to the start of the value (supporting reading to the end), or
- relative to the end of the value (supporting reading to the start).

Additionally, the details of how a byte range should be implemented (e.g. enum or negative offset/size) should be left up to the implementation.

PR #308, opened 2024-08-21
clarify that the abstract store api is not normative

### LINKED ISSUE #307: Abstract store interface should be clarified as descriptive, not normative
I would like to understand why the spec includes a [Abstract store interface](https://zarr-specs.readthedocs.io/en/latest/v3/core/v3.0.html#abstract-store-interface) section and ask if this section should be clarified as descriptive, not normative.

Why should the spec care what interface and implementation chooses for the Store abstraction? So long as the store does the right thing, surely the implementation is free to choose a different abstract interface?

jhamman:
closes #307

### DIFF
--- a/docs/v3/core/v3.0.rst
+++ b/docs/v3/core/v3.0.rst
@@ -1271,7 +1271,7 @@ when opening a Zarr hierarchy to automatically determine the appropriate store.
 Abstract store interface
 ------------------------
 
-(This subsection is not normative.)
+(This subsection is not normative. It is intended only as a recommendation for API design, but does not need to be strictly followed like the specification for the on-disk storage does.)
 
 The store interface is intended to be simple to implement using a
 variety of different underlying storage technologies. It is defined in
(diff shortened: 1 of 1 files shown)


LDeakin:
This would also close #266

jbms (dismissed):

docs/v3/core/v3.0.rst
```diff
@@ -1273,6 +1271,8 @@ when opening a Zarr hierarchy to automatically determine the appropriate store.
 Abstract store interface
 ------------------------
 
+(This subsection is not normative.)
```
TomNicholas, line 1274: I feel like this could be a lot clearer, e.g:

```suggestion
(This subsection is not normative. It is intended only as a recommendation for API design, but does not need to be strictly followed like the specification for the on-disk storage does.)
```

rabernat (approved):

joshmoore:
Tried relaunching the RTD build.

LDeakin:
This keeps coming up, so it'd be good to get this merged

d-v-b (approved):

LDeakin (approved):

Issue #103, opened 2020-10-23
Non-deletable stores. 


Carreau:
I just realized that the zipStore does not allow deletion (and thus does not allow rename). 
Is that something that should be addressed in the spec, of should it maybe be fixed in the ZipStore with TombStones ?

alimanfoo:
Hi @Carreau, I think it is reasonable if a store does not support deletion. If you agree, then yes I'd suggest we clarify that supporting deletion is optional somehow. E.g., maybe stores can be readable, writable, deletable, listable, these are the basic capabilities. A store that is only readable can still be useful. Similarly a store that is only readable and writable can still be useful.

Carreau:
That's fine with me, but that's push more complexity on the Zarr Protocol implementation, while you could try to push the deletion implementation details to the store. That seem also to be there only to accommodate the zip store.

jakirkham:
Think what Matthias was referring to is `ZipStore` allows writing, but not deletion. So the `erase*` functions don't work. Similarly writing behaves more like write-once (since deletion is not supported).

That said, think in practice `ZipStore` behaves better as a "readable store" (ignoring the initial creation step). This is already a common case like a read-only filesystem or HTTP stores is a common use case that we would want to support, which is covered under "readable store".

Side note: We may prefer to fix the `ZipStore` oddity ( https://github.com/zarr-developers/zarr-python/issues/828 ) either by making it more feature complete ( https://github.com/zarr-developers/zarr-python/pull/1184 ) or moving it to read-only.

jstriebel:
Thanks for the clarification @jakirkham! Do you think there's a TODO for the current v3 core spec from this issue? I'm re-opening it, but I'm not quite sure if this belongs into the zarr-specs repo or rather zarr-python? Sorry for the confusion, seems like I'm missing some context here. @joshmoore, can you help?

joshmoore:
Seems like there's enough confusion around this that we should minimally try to add our expectations in the V3 spec and possibly V2. @Caviere will soon be investigating the ZipStore and we can try to dig into these questions which may then lead to wanting to address delete-ability more broadly in the spec.

jstriebel:
Since the zip store is currently not yet specified in v3 I assume this does not block ZEP 1, which is mostly about the core spec. Further clarifications can also be added to the core spec after ZEP 1 is accepted.

PR #311, opened 2024-09-11
feature(stores): draft zip file store specification

jhamman:
This is a working draft of the v3 ZIP file store specification. 

xref: 

- https://github.com/zarr-developers/zarr-python/pull/2078
- https://github.com/zarr-developers/zarr-specs/issues/103

### DIFF
--- a/docs/v3/stores/zipfile/v1.0.rst
+++ b/docs/v3/stores/zipfile/v1.0.rst
@@ -1,0 +1,152 @@
+.. _zip-file-store-v1:
+
+=============================
+ ZIP file store (version 1.0)
+=============================
+
+Specification URI:
+    https://zarr-specs.readthedocs.io/en/latest/v3/stores/zipfile/v1.0.html
+Corresponding ZEP:
+    `ZEP0001 — Zarr specification version 3 <https://zarr.dev/zeps/accepted/TODO.html>`_
+Issue tracking:
+    `GitHub issues <https://github.com/zarr-developers/zarr-specs/labels/stores-filesystem-v1.0>`_
+Suggest an edit for this spec:
+    `GitHub editor <https://github.com/zarr-developers/zarr-specs/blob/main/docs/v3/stores/zipfile/v1.0.rst>`_
+
+Copyright 2024-Present Zarr core development team. This work is
+licensed under a `Creative Commons Attribution 3.0 Unported License
+<https://creativecommons.org/licenses/by/3.0/>`_.
+
+----
+
+
+Abstract
+========
+
+The ZIP file store provides a way to store Zarr arrays and metadata in a
+single, compressed ZIP archive. This specification defines the organization
+and structure of the ZIP file to ensure compatibility across implementations.
+
+Status of this document
+=======================
+
+This is a working draft.
+
+Notes about design decisions for the native ZIP File Store 
+==========================================================
+
+The ZIP file store is designed for simplicity and easy conversion to and from
+the filesystem stores. The ability to ZIP a file system store using standard
+command line tools (e.g. `zip`) is a key virtue of the design below.
+
+Document conventions
+====================
+
+Conformance requirements are expressed with a combination of
+descriptive assertions and [RFC2119]_ terminology. The key words
+"MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in the normative
+parts of this document are to be interpreted as described in
+[RFC2119]_. However, for readability, these words do not appear in all
+uppercase letters in this specification.
+
+All of the text of this specification is normative except sections
+explicitly marked as non-normative, examples, and notes. Examples in
+this specification are introduced with the words "for example".
+
+
+Native storage operations
+=========================
+
+Here we consider a ZIP file to be a standard ZIP archive, where:
+
+* Each key has a name (sequence of characters) and contents
+  (sequence of bytes).
+
+* Each directory has a name (sequence of characters) and children (set
+  of zero or more files and/or directories).
+
+* Each file or directory can be addressed by a path, comprised of its
+  name and the names of all ancestor directories, which uniquely
+  identifies it within the file system.
+
+… and where the following native operations are supported:
+
+* Create a file.
+
+* Write the contents of a file.
+
+* Read the contents of a file.
+
+* Create a directory.
+
+* List the children of a directory, returning the name and type (file
+  or directory) of each child.
+
+… Note that the following operations may not be supported by ZIP file stores:
+
+* Delete a file.
+
+* Delete a directory.
+
+Key translation
+===============
+
+The Zarr store interface is defined in terms of `keys` and `values`,
+where a `key` is a sequence of characters and a `value` is a sequence
+of bytes. A ZIP file store represents keys as paths within a ZIP 
+archive. No further translation of keys is required.
+
+Store API implementation
+========================
+
+The section below defines an implementation of the Zarr
+:ref:`abstract-store-interface` in terms of the native operations of this
+storage system.
+
+* ``get(key) -> value`` : Read and return the contents of the object at
+  within the archive at path ``key``. 
+
+* ``set(key, value)`` : Write ``value`` as the contents of the file at
+  into the archive at path ``key .
+
+* ``list()`` : List all keys in the archive.
+
+* ``list_prefix(prefix)`` : List all keys within the archive that begin
+  with ``prefix``.
+
+* ``list_dir(prefix)`` : List all keys within a directory within the archive.
+
+
+Canonical URI
+=============
+
+The canonical URI format for this store follows the file URI scheme of the
+archive itself, as defined in [RFC8089]_. For a Windows base directory path
+"c:\\my data.zip" the canonical URI would be "file:///c:/my%20data.zip", 
+for a Posix base directory "/my data.zip" it would be"file:///my%20data.zip".
+
+
+Store limitations
+=================
+
+The following limitations for this store are know:
+
+* ZIP file stores may not implement delete or rename operations
+
+
+References
+==========
+
+.. [RFC2119] S. Bradner. Key words for use in RFCs to Indicate
+   Requirement Levels. March 1997. Best Current Practice. URL:
+   https://tools.ietf.org/html/rfc2119
+
+.. [RFC8089] M. Kerwin. The "file" URI Scheme. February 2017. Proposed Standard.
+   URL: https://tools.ietf.org/html/rfc8089
+
+
+Change log
+==========
+
+@@TODO


docs/v3/stores/zipfile/v1.0.rst
```diff
@@ -0,0 +1,152 @@
...
+* Create a directory.
+
+* List the children of a directory, returning the name and type (file
+  or directory) of each child.
+
+… Note that the following operations may not be supported by ZIP file stores:
+
+* Delete a file.
+
+* Delete a directory.
```
jhamman, line 90: See https://github.com/zarr-developers/zarr-specs/issues/103

joshmoore:
In my experience, the root of the zip is one of the trickiest parts for data creators (and I assume implementers) to get right, e.g.,

 * https://github.com/zarr-developers/zarr-python/issues/756#issuecomment-852045828
 * https://github.com/zarr-developers/zarr-python/commit/755bb961381a53569f6826cbc230bbc99e02a03a

joshmoore:
cc: @DennisHeimbigner

zoj613:
How useful is a ZipStore in practice? Are there a lot of use cases for it? Given how limited it is (no rename/deletion, etc) I am wondering if its worth having a spec for it

DennisHeimbigner:
I have support equivalent to zipstore in nczarr in the netcdf-c library. I agree that it does not appear to be
very useful, but the basic idea behind it is reasonable: a single file containing a complete zarr file tree,
and using compression the component files to save space.
Personally, I think that using a single file file system (SFFS) with added compression makes more sense.
There are several implementations available, and it is easy enough to write your own,

jhamman:
> In my experience, the root of the zip is one of the trickiest parts for data creators (and I assume implementers) to get right...

@joshmoore - do you have suggestions for the spec document that would make this clearer?

---

@zoj613 and @DennisHeimbigner - let's try to avoid making this about alternatives to the ZIP store concept. There are practical reasons to add this (Zarr-Python has long supported a ZIP store interface). 

Remember, Zarr can support many storage backends. If there are alternatives to experiment with, let's do that in a separate issue.

docs/v3/stores/zipfile/v1.0.rst
```diff
@@ -0,0 +1,152 @@
...
+
+The section below defines an implementation of the Zarr
+:ref:`abstract-store-interface` in terms of the native operations of this
+storage system.
+
+* ``get(key) -> value`` : Read and return the contents of the object at
+  within the archive at path ``key``. 
+
+* ``set(key, value)`` : Write ``value`` as the contents of the file at
+  into the archive at path ``key .
```
zoj613, line 111: Is the use of `at within` and `at into` in these lines intentional? Sounds like a typo

docs/v3/stores/zipfile/v1.0.rst
```diff
@@ -0,0 +1,152 @@
...
+this specification are introduced with the words "for example".
+
+
+Native storage operations
+=========================
+
+Here we consider a ZIP file to be a standard ZIP archive, where:
+
+* Each key has a name (sequence of characters) and contents
+  (sequence of bytes).
```
jhamman, line 64: Note that the keys are relative paths (not prefixed with a `/`).

DennisHeimbigner:
> for the format, the most important item I know of is "don't include the top-level directory" (though I have run into some complaints about that from various repositories, since the behavior differs between implementations, e.g. on Windows)

I think I have always used either linux zip or cygwin zip to create zarr zip files. What native windows program could I use to create a pure windows zip file?
As for the top-level directory, I think it is better to always include it. I say this so that my rule holds, namely:
1.unzipping a zip store creates a directory tree usable by the zarr directory tree storage manager.
2. zipping a zarr directory tree creates a zip store conforming to the proposed zip spec.

jhamman:
> I think I have always used either linux zip or cygwin zip to create zarr zip files. What native windows program could I use to create a pure windows zip file?

🤷 

>As for the top-level directory, I think it is better to always include it. I say this so that my rule holds, namely:
> 1.unzipping a zip store creates a directory tree usable by the zarr directory tree storage manager.
> 2. zipping a zarr directory tree creates a zip store conforming to the proposed zip spec.

👍

joshmoore:
A few downsides of adding the directory:
 - It's a change from v2 :smile: though that can be handled.
 - You must know/lookup that value before you open the zip.
 - What is the behavior if there are multiple .zarr directories within the zip?

docs/v3/stores/zipfile/v1.0.rst
```diff
@@ -0,0 +1,152 @@
...
+The canonical URI format for this store follows the file URI scheme of the
+archive itself, as defined in [RFC8089]_. For a Windows base directory path
+"c:\\my data.zip" the canonical URI would be "file:///c:/my%20data.zip", 
+for a Posix base directory "/my data.zip" it would be"file:///my%20data.zip".
+
+
+Store limitations
+=================
+
+The following limitations for this store are know:
```
d-v-b, line 133:
```suggestion
The following limitations for this store are known:
```

d-v-b:
thanks for the writeup @jwindhager et al. It might be clarifying to factor out the following two discussions, as I think they are logically separable:
- question 1: how should Zarr implementations design APIs for the storage of Zarr data inside single-file archive formats (namely zip)? I think this is the question @jhamman is attempting to resolve with this particular PR. I think it may not be possible to answer this particular question in a way that provides _normative_ guidelines for applications that target end users.
- question 2:  can we define best practices (or even a spec) for storing Zarr data inside single-file archives (namely, zip)? I think this is what @jwindhager  et al are looking for (please correct me if I'm misreading). Such best practices / spec might reasonably carve out a restricted subset of the space expressible with a generic Zarr + zip API, and include things like file extensions, URI conventions, etc. 
 
I think question 2 is the kind of thing that really ought to be explored alongside at least one implementation. So my pitch is as follows: we _implement_ an opinionated zarr archive function in `zarr-python`.  The function I am imagining takes a zarr array or group and saves all the stuff in that array / group in a destination zip file, according to a set of guidelines that we simultaneously document as a spec-like-document in the `zarr-python` repo. I think this could be really productive, and developing the spec alongside the implementation will keep them both balanced. At the end of this process, we should have both a useful implementation and a well-tested spec, which could be re-published in a broader venue than `zarr-python`.

Would anyone here be interested in working on such an effort?  I do think this requires at least one champion to push forward. As a `zarr-python` maintainer I would be willing to put in the time for reviewing PRs and whatnot.

jbms:
I'm in favor of supporting archive formats and zip.  Tensorstore already supports reading but not writing. 

Zip has some disadvantages in its design but I think they are outweighed by it being such a common format. 

I agree that there should be no implicit root directory, and while some implementations may do auto-discovery, there should be a canonical url that makes any sub-directories within the zip file explicit.

The spec says the canonical url is just a file url, file:///path/to/file.zip.  While that is reasonable for implementations that do auto-discovery, I don't think that is a good idea as the canonical url since it does not explicitly indicate the zip format at all, and would rely on implementations detecting it either by the filename or content.

Previously I proposed a different url syntax (https://github.com/zarr-developers/zeps/pull/48) which allows nested formats like zip to be specified explicitly.

DennisHeimbigner:
IMO zip is not a good single file storage format. Other choices like the various
single file file systems (SFFS) seem to me to be a much more flexible choice, especially
in the face of writing.

d-v-b:
> IMO zip is not a good single file storage format. Other choices like the various
> single file file systems (SFFS) seem to me to be a much more flexible choice, especially
> in the face of writing.

This might be true but I think it's orthogonal to the discussion at hand -- we are not trying to find the best archive file format, but rather devise standards to improve the utility of a popular archive file format (zip). Zip being sub-par doesn't bear on the fact that that people want to use it, and that latter fact is what we should build around IMO.

zoj613:
> The current draft does not specify that existing keys cannot generally be overwritten (to our understanding, this is not generally possible according to the ZIP standard).

I think this should be left to the implementation to decide whether to support write operations. Some people might want to rename/overwrite/delete entries in a ZIP as a convenience just like any other store. Sure the ZIP standard does not support this but there are ways to workaround this limitation (although quite inefficient). For example, I have 2 ZipStore implementations that support the full zarr v3 abstract store interface as defined [in the core spec](https://zarr-specs.readthedocs.io/en/latest/v3/core/v3.0.html#abstract-store-interface). I don't see the benefit of imposing this limitation to implementations.

jwindhager:
@d-v-b I agree that we *could* factor out the two discussions (this is what we meant with "separate the interface definition from the on-disk representation") and I would also add the "canonical url" proposal by @jbms to the list (this is roughly what we referred to as "consistent resource identifier schemes"). However, given the way stores are specified in the current spec, I'd pragmatically argue that it's probably easier to get this PR merged in its current form rather than to further broaden the scope / branch out.

I also agree that we could push the on-disk representation aspect on the `zarr-python` side ("guidelines"), but we'd need to make sure to simultaneously push this on the spec side (i.e. here) as well. Otherwise, we may risk ending up with a Python-specific de-facto standard that in the worst case collides with future upstream specs. But perhaps this worry is unfounded, in which case I'd be happy to contribute within my (unfortunately limited) time constraints.

---

@jbms Thanks a lot for pointing us to your ZEP! We somehow completely missed it in our group's discussions. I think this could address many important issues, also related to zipped Zarr (and thus single-file OME-NGFF), and will give this a read asap.

---

> > > The current draft does not specify that existing keys cannot generally be overwritten (to our understanding, this is not generally possible according to the ZIP standard).
>>
> > I think this should be left to the implementation to decide whether to support write operations. Some people might want to rename/overwrite/delete entries in a ZIP as a convenience just like any other store. Sure the ZIP standard does not support this but there are ways to workaround this limitation (although quite inefficient). For example, I have 2 ZipStore implementations that support the full zarr v3 abstract store interface as defined [in the core spec](https://zarr-specs.readthedocs.io/en/latest/v3/core/v3.0.html#abstract-store-interface). I don't see the benefit of imposing this limitation to implementations.
>
> Agreed, and in fact I think this spec be made much more concise. I don't think it is necessary to list the supported operations.

Fair point. I still think that one could clarify this under "Store limitations" using the right phraseology, but no strong feelings.

d-v-b:
> I also agree that we could push the on-disk representation aspect on the `zarr-python` side ("guidelines"), but we'd need to make sure to simultaneously push this on the spec side (i.e. here) as well. Otherwise, we may risk ending up with a Python-specific de-facto standard that in the worst case collides with future upstream specs. But perhaps this worry is unfounded, in which case I'd be happy to contribute within my (unfortunately limited) time constraints.

We can iterate quickly in `zarr-python`, _including_ on the spec side. What I'm imagining is "incubating" the zip convention in `zarr-python` (and / or any other zarr implementation), developing a spec and an implementation that are both tested, and finally propose something to this repo with that experience in hand. Until the spec is published here, we would publish it in zarr-python, and the spec itself would not be python-specific, because we would design it that way :)

auxym:
In addition to not supporting deletion, zip storage would also have the following issues with `append()`:

1. The metadata in the array json needs to be updated, this is not trivial in a zip file
2. If the append is not a full chunk (or multiple full chunks) or the current shape of the array is not an exact multiple of the chunk size, then the chunk files (in the zip) need to to be appended to (I think?) which once again is not supported by the zip format.

The current implementation of ZipStorage in zarr-python seems to hit the first issue when attempting to append (tries to re-write zarr.json, error due to an existing file of the same name):

```
In [69]: zs = zarr.storage.ZipStore("test2.zarr", mode="w")

In [70]: root = zarr.create_group(store=zs)

In [71]: a1 = root.create_array("a1", shape=(1000,), chunks=(1000,), dtype=float)

In [72]: a1.append(np.arange(1000, dtype=np.float64))
C:\Users\thef2102\AppData\Roaming\uv\python\cpython-3.12.10-windows-x86_64-none\Lib\zipfile\__init__.py:1611: UserWarning: Duplicate name: 'a1/zarr.json'
  return self._open_to_write(zinfo, force_zip64=force_zip64)
Out[72]: (2000,)
```

Should the spec specify that arrays in zip stores are write-once only, so in addition to not being eraseable, they are not resizeable or append-able?

I can see some ways to work around the issues and support appends, but they might be inefficient, very complex and/or leaky abstractions that make the storage not entirely opaque to code users.

BTW, I've written a comment a #209, since it seems this issue is not the place to discuss alternatives to zip, but I think sqlite would be a good candidate for single-file zarr stores that support updates, appends and deletions.

d-v-b:
i am far from a zip expert but my understanding is that you can modify the contents of a zip archive, at the price of increasing its size. This means one can appending to zipped zarr data, e.g., by adding new arrays or groups, but one could also modify existing objects in the archive. Evidently we don't support this in zarr python but that seems like a problem with our implementation, not the zip format.

zoj613:
> Another thought on this topic.
> 
> The Store API, according to the zarr spec, currently defines 3 optional "properties" of a store:
> 
>     * Readable: supports get, get_partial operations
> 
>     * Listable: Supports list, list_prefix, list_dir operations
> 
>     * Writable: Supports set, erase, etc.
> 
> 
> Perhaps it would be relevant to add a addition property named "editable" or, conversely, "append-only"? An append-only store would support `set` _only if_ the key does not already exist, and would _not_ support `delete`. This describes the current implementation of ZipStore in zarr-python and makes sense generally for the zip file format. I think it could apply to other eventual store implementations, too.

I feel what you're describing is just an implementation detail and should really be up to package maintainers to decide (As long as it is communicated in the documentation). Adding an "editable" store classification brings nothing new and is already covered by the 3 store classes.

mkitti:
xref: https://ngff.openmicroscopy.org/rfc/9/index.html

dblodgett-usgs:
This issue has been languishing. Should probably be closed or accepted?

I recently implemented a Zarr v3 client, and a specified zip store would have made test fixtures considerably easier and more portable.

My read is that each open item could go either way and be workable. Toward that end, here's how I'd resolve them.

**No root directory.** `unzip -d my.zarr archive.zip` gives the directory-store round-trip @DennisHeimbigner was looking for without requiring the archive to carry a name the reader has to discover before opening it. It also removes the ambiguity @joshmoore raised about archives containing more than one `.zarr` directory.

**Stored (method 0) as the recommended default, other methods allowed.** Zarr chunks already carry their own codecs, so deflating them again mostly costs CPU for little gain. Readers MUST support stored and deflate; writers SHOULD default to stored.

**Canonical URI as a follow-up.** @jbms is right that a bare `file:///path/to/file.zip` doesn't identify the format, but nested-format URL syntax (zeps#48) is a larger design than this store spec and I'd rather not see the two coupled. The store spec can define the on-disk layout and leave identification to a separate document.

**Mutability — no position.** The abstract store interface already distinguishes readable/listable/writable. I don't think this spec needs to say more, and I'd rather it didn't block on the question.

Not trying to reopen alternatives to zip here — separate discussion, as noted upthread.

Can we agree to a workable zip store convention and move forward here?

d-v-b:
as the contents of a zip archive can be directly interpreted as files and directories, and addressed internally with conventional paths, I'm not sure how much normative content we need to guarantee interoperability among implementations. usage guidelines seem helpful but are there ambiguities that require a full spec to resolve?

dblodgett-usgs:
Perhaps not. Acknowledging the validity of the approach in a normative statement would help herd sheep and get to some kind of consistency though, right?

mkitti:
Note the RFC specification at https://ngff.openmicroscopy.org/rfc/9/index.html

dblodgett-usgs:
Thanks for that. Is there a reason not to have a zarr3 zip store that follows the lead of OME-Zarr?

mkitti:
There should probably be a generic specification for Zarr in zip. Part of the motivation of pursuing it in OME rather than here was partly because this effort stalled.

There are a few additional aspects I have identified.

1. The Python standard library zipfile needs improvement for an effective implementation in zarr-python. I particular the ability to remove entries in the central directory and repack the archive. Those improvements landed a few weeks ago after about 17 years of consideration: https://github.com/python/cpython/commit/aec0aed197872adf96d3f25fd5ea624508fcddbf

2. There is a lack of clarity of some Zarr concepts such as the notion of a "root" that sorely needs to be clarified.

Right now, I think the best thing would be let the OME specification continue to evolve. Meanwhile, I think it is worth considering if those recommendations generalize well.

In particular, do we still want to pursue this solely as a single file hierarchy or could we have a constellation of zip files with sub-hierarchies? If so, how are those mounted to their parent hierarchy and how do we define the root of that hierarchy?

Also what is the role of the specification? There seems to be a gap between aspects that implementations could probably tolerate and what would make implementation in a remote cloud store optimal, for example zip-level compression. In constructing the OME specification we found that sharding within a zip file is important for scalability of the format. The zip central directory is problematic for a larger number of individual chunk files. The solution is to use sharding within the zip file and use the shard index rather the zip central directory to index individual small (inner) chunks. To access the shard index and inner chunks effectively in the cloud, the shard itself should not be compressed since that would compromise the ability to do partial reads within the shard.

I'm also coming to the thought that a native Zarr archive might be worth pursuing. The shard index concept could be extended to index strings or even files. Unlike the zip central archive which has variable length records, a fixed length record of a shard index makes random access a lot easier. Rather than working around the features of other formats perhaps we should extend existing concepts that work well for us.

dblodgett-usgs:
> There should probably be a generic specification for Zarr in zip.

I agree. Working with Claude to reconcile these efforts... something along the lines of: 

- **Key mapping: zip entry names are store keys verbatim, archive root == store root**
  - Two implementations that disagree here can't read each other's archives; nothing above the store layer can fix it.
- **Root disambiguation: separate "store root" from "hierarchy root"**
  - @mkitti's point 2 — the ambiguity is in Zarr core, not in OME, so a domain profile can't resolve it.
- **Central directory authoritative; local file headers not trusted; duplicates last-wins**
  - Silent-wrong-answer failure mode, and current zarr-python + xarray writes trigger it today.
- **Entries MUST be range-readable; shard entries MUST be stored (method 0)**
  - Compressing a shard destroys shard-index partial reads — true of any sharded Zarr in zip, not just images.
- **Zip-level compression allowed, SHOULD default to stored; readers MUST support stored and deflate**
  - Interop floor. Only the core spec can say what every reader must decode.
- **Store capability mapping: writes / deletes / partial writes / listing, plus repack semantics**
  - Belongs with the abstract store interface it maps onto; the CPython entry-removal change makes this newly answerable.
- **Cloud-optimal profile (SHOULD): ZIP64, sharding, metadata-first ordering, no multi-volume**
  - Generic performance properties of zip-over-object-storage; OME rediscovered them, and the next domain shouldn't have to.
- **Non-goals: file extension, archive-comment metadata, nested-archive mounting and URI syntax**
  - Domain-specific or separate designs — naming them keeps scope from reopening every review round.
  
  @jhamman -- are you still hoping to push this PR to consensus? Would you be willing to update in the interest of a base zip store convention that is consistent with practice that's emerged since you drafted?

d-v-b:
> Entries MUST be range-readable

i'm a bit skeptical of any zip spec that would require zarr implementations to obey rules that are not absolutely required for interoperability. I don't think range-read support is required for interoperability. It's a _convenience_ for access over http, and maybe it's so convenient that some application like ome-zarr makes it a requirement. But for core zarr implementations, "range readable" is _not_ required for interop, because an implementation can just read the whole thing.

what exactly in the proposed zip spec would be required for interoperability? And if the answer is "nothing", then maybe we can think about a non-spec framing for the (very useful!) guidelines for storing zarr data in zip archives.

dblodgett-usgs:
Fair point. From one perspective (considering the whole archive) nailing down the key -> value mapping isn't an issue. I think it would be fine for the base spec to be scoped in a way that a zip store MAY have to be treated as a monolith (not be range-readable). 

Refining the above:

**Normative — required for interoperability**

- **Entry-name-to-key mapping, including archive root == store root**
- **Duplicate-entry resolution: central directory authoritative, last entry wins**
- **Readers MUST support stored and deflate; writers MAY use any [APPNOTE](https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT)-registered method**
- **Multi-volume and encrypted archives are note supported**

**Non-normative — guidance, not conformance**

- **Range-readability of entries**
- **Chunk and shard entries stored (method 0); metadata entries MAY be deflated**
- **ZIP64 handling**
- **Sharding and metadata-first entry ordering**
- **Store capability mapping and repack semantics**

**Out of scope**

- **File extension, archive-comment metadata, nested-archive mounting and URI syntax**

Issue #316, opened 2024-10-11
Lessons to learn from STAC's extensibility


TomAugspurger:
As mentioned in https://github.com/zarr-developers/zarr-specs/pull/309, I ran across some challenges with how the Zarr v3 spec does extensions. I think that we might be able to learn some lessons from how [STAC](https://github.com/radiantearth/stac-spec/) handles extensions.

---

tl/dr: I think Zarr would benefit from a better extension story that removed the need to have any involvement from anyone other than the extension author and any tooling wishing to use that extension. JSON schema + a `zarr_extensions` field on `Group` and `Array` would get us most of the way there. The current requirements of `must_understand: false` and `name: URL` in the extension objects feels like a weaker version of this.

---

## How STAC does extensibility

STAC is a JSON-based format for cataloging geospatial assets. https://github.com/radiantearth/stac-spec/blob/master/extensions/README.md#overview lays out how STAC allows itself to be extended, but there are a few key components

1. STAC uses [jsonschema](https://www.jsonschema.net) to define schemes for both the core metadata and extensions.
2. All STAC objects (`Collection`, `Item`, etc.) include a `stac_version` field.
3. All STAC objects (`Collection`, `Item`) include a `stac_extensions` array with a list of URLs to JSON Schema definitions that can be used for validation.

Together, these are sufficient to allow extensions to extend basically any part of STAC without any involvement from the core of STAC. Tooling built around STAC coordinates through `stac_extensions` For example, a validator can load the JSON schema definitions for the core metadata (using the `stac_version` field) and all extensions (using the URLs in `stac_extensions`) and validate a document against those schemas. Libraries wishing to use some feature can check for the presence of a specific `stac_extension` URL.

You also get the ability to version things separately. The core metadata can be at `1.0.0`, while the `proj` extension is a 2.0.0 without issue.

## How that might apply to Zarr

Two immediate reactions to the thought of applying that to Zarr:

1. Zarr does have JSON documents for describing the metadata of nodes in a Zarr hierarchy. We could pretty easily take the same concepts and apply them more or less directly to the `Group` and `Array` definitions (and possibly other fields within; STAC does this as well for, e.g. `Assets` which live inside an `Item`).
2. STAC is entirely JSON-based, while much of Zarr concerns how binary blobs are stored, transformed, etc. While portions of these extension points might be configured (and validated by JSON schema) in the metadata document, much of it will lie outside.

## How does this relate to what zarr has today?

[...]

PR #317, opened 2024-10-11
Clarify extra / must_understand language

TomAugspurger:
xref https://github.com/zarr-developers/zarr-specs/issues/316#issuecomment-2407162818: clarifies that extra keys are allowed in the node metadata, they just must have this `must_understand` field set to `false`.


d-v-b (approved):

d-v-b:
thanks!

rabernat (approved):
I don't know why we let this languish so long. Seems like a minor clarification, should be merged asap.

PR #320, opened 2024-12-06
Correct example endianness reference

thewtex:
The examples references a "configuration below" as little endian, but the configuration contains:

```json
        "codecs": [{
            "name": "bytes",
            "configuration": {
                "endian": "big"
            }
        }],
```

I assume that this is the right fix. If not, it would be helpful to extend the documentation to provide a more detailed explanation.

In general, it would be helpful to document how endianness works with Zarr 3.

I currently see just the note:

> We are explicitly looking for more feedback and prototypes of code using the r*, raw bits, for various endianness and whether the spec could be made clearer.

And the [Data Types](https://zarr-specs.readthedocs.io/en/latest/v3/data-types.html) page is empty.

@jbms @normanrz @joshmoore @MSanKeys963

### DIFF
--- a/docs/v3/core/v3.0.rst
+++ b/docs/v3/core/v3.0.rst
@@ -720,7 +720,7 @@ compressed using gzip compression prior to storage::
         "codecs": [{
             "name": "bytes",
             "configuration": {
-                "endian": "big"
+                "endian": "little"
             }
         }],
         "fill_value": "NaN",


normanrz:
Thanks for pointing this out. We could also change the example to `"endian": "little"`, no?

thewtex:
> We could also change the example to "endian": "little", no?

Yes, that would work, too. Is that preferred?

thewtex:
> I think I would use little endian in the example because it is much more common these days. No strong opinion here, though.

:+1: flipped to little endian.

joshmoore:
Re-launched the RTD build. If it keeps failing, then something has likely broken upstream.

joshmoore:
Unrelated to this PR:

```
  File "/home/docs/checkouts/readthedocs.org/user_builds/zarr-specs/envs/320/lib/python3.11/site-packages/packaging/version.py", line 202, in __init__
    raise InvalidVersion(f"Invalid version: {version!r}")
packaging.version.InvalidVersion: Invalid version: ''
```

jbms (approved):

PR #326, opened 2025-01-21
Remove inconsistent example description

tpietzsch:
There is nothing in the example specifying compression.

I'm also not sure about the sentence "Within each chunk the binary values are laid out in C contiguous order", because this is the implicit default and does not correspond to anything in the metadata example? Maybe add something about endianness instead?


joshmoore:
Pushed an attempted fix for the build.

jbms (approved):

PR #333, opened 2025-02-23
Update statuses

### LINKED ISSUE #332: The v2 specs state that it's the latest
From [the v2 specs](https://zarr-specs.readthedocs.io/en/latest/v2/v2.0.html):

> This specification is the latest version. See [v3](https://zarr-specs.readthedocs.io/en/latest/specs.html#zarr-specs) for previous versions.

Is v3 before v2?

sanketverma1704:
Fixes #332.

### DIFF
--- a/docs/v1/v1.0.rst
+++ b/docs/v1/v1.0.rst
@@ -13,9 +13,7 @@ interpreted as described in `RFC 2119
 Status
 ------
 
-This specification is deprecated. See `v3
-<https://zarr-specs.readthedocs.io/en/latest/v3/core/v3.0.html>`_ for the
-latest version.
+This specification is deprecated. See :ref:`zarr-specs` for the latest version.
 
 Storage
 -------
--- a/docs/v1/v1.0.rst
+++ b/docs/v1/v1.0.rst
@@ -13,7 +13,9 @@ interpreted as described in `RFC 2119
 Status
 ------
 
-This specification is deprecated. See :ref:`zarr-specs` for the latest version.
+This specification is deprecated. See `v3
+<https://zarr-specs.readthedocs.io/en/latest/v3/core/v3.0.html>`_ for the
+latest version.
 
 Storage
 -------
--- a/docs/v2/v2.0.rst
+++ b/docs/v2/v2.0.rst
@@ -12,8 +12,8 @@ used for storing Zarr arrays. The key words "MUST", "MUST NOT", "REQUIRED",
 Status
 ------
 
-This version is outdated and not actively maintained. See :ref:`zarr-specs` for
-the latest version.
+This specification has been superseded. See :ref:`zarr-specs` for the latest
+version.
 
 .. _spec_v2_storage:
 
--- a/docs/v2/v2.0.rst
+++ b/docs/v2/v2.0.rst
@@ -12,9 +12,8 @@ used for storing Zarr arrays. The key words "MUST", "MUST NOT", "REQUIRED",
 Status
 ------
 
-This version is outdated and not actively maintained. See `v3
-<https://zarr-specs.readthedocs.io/en/latest/v3/core/v3.0.html>`_ for the
-latest version.
+This version is outdated and not actively maintained. See :ref:`zarr-specs` for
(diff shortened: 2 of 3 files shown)


d-v-b:
Why change the text to refer to the newer version, when previously it was referring to the older version (zarr v1)?

sanketverma1704:
Previously, the link in the `STATUS` section across v1 and v2 referred to this page: https://zarr-specs.readthedocs.io/en/latest/specs.html.

I've updated the `STATUS` section on all three versions, linking to the latest version.

docs/v1/v1.0.rst
```diff
@@ -13,7 +13,9 @@ interpreted as described in `RFC 2119
 Status
 ------
 
-This specification is deprecated. See :ref:`zarr-specs` for the latest version.
+This specification is deprecated. See `v3
+<https://zarr-specs.readthedocs.io/en/latest/v3/core/v3.0.html>`_ for the
```
d-v-b, line 17: with this change, we will need to update the 1.0 spec again if we ever release a version after 3

sanketverma1704: Should we remove the link for v3 from v1 and v2, then?

docs/v3/core/v3.0.rst
```diff
@@ -40,6 +40,8 @@ Status of this document
 
 ZEP0001 was accepted on May 15th, 2023 via https://github.com/zarr-developers/zarr-specs/issues/227.
 
+This specification is the latest version.
```
d-v-b, line 43:
```suggestion
```
If the spec says "this is the latest version", then we have to change the spec when we release a new version. We can prevent this by simply not stating "this is the latest version".

docs/v1/v1.0.rst
d-v-b: It is true forever for both v1 and v2 that they have been superseded, but it doesn't make sense to hard-code the latest version of the spec in the old spec documents. I think the correct thing is to say "there's a new version [link to list of zarr spec versions]", which I think is what was supposed to happen from the start

d-v-b:
to keep things simple, how about just fix the specific problem reported in https://github.com/zarr-developers/zarr-specs/issues/332?

docs/v3/core/v3.0.rst
sanketverma1704: I agree, but I think the author of the latest version should take responsibility for adding the text to avoid confusion, as seen in #332.

d-v-b:
the problem reported in #332 is that the link text is "v3" when it should be "the list of zarr specs" or something equivalent

docs/v2/v2.0.rst
```diff
@@ -12,8 +12,8 @@ used for storing Zarr arrays. The key words "MUST", "MUST NOT", "REQUIRED",
 Status
 ------
 
-This specification is the latest version. See :ref:`zarr-specs` for previous
-versions.
+This version is outdated and not actively maintained. See :ref:`zarr-specs` for
```
sanketverma1704, line 15: Is 'This version is outdated and not actively maintained' good?

d-v-b: I don't think we need to change this text here. let's keep the changes to the absolute minimum required to address the linked issue.

sanketverma1704: I'm sorry, I don't understand.

Not changing the text to say, 'This specification is the latest version' in the V2 spec doesn't solve the issue.

d-v-b: oops, I was confused about which file this was, sorry about that. I recommend saying "This specification has been superseded. See :ref:`zarr-specs` for the latest version.". I now see that it doesn't make sense to keep the original reference to previous versions, but just saying "there's a newer version" means nobody has to touch this text again if we release v4.

sanketverma1704: 'This specification has been superseded.' is a better statement.
Thanks, @d-v-b.

d-v-b (approved):

jbms (approved):

PR #347, opened 2025-06-02
Added json schema

TomAugspurger:
This adds a pair of [json schema](https://json-schema.org/docs) schemas to the repository. One for Array metadata and one for Group metadata.

For those unfamiliar with json-schema, it's a language for validating JSON documents. You write schemas (in JSON) and tools can validate JSON objects ("instances") against that schema. For example, the following Group would be flagged as invalid, because it lacks a `zarr_format` field:

```json
{
    "node_type": "group",
    "attributes": {
        "spam": "ham",
        "eggs": 42
    }
} 
```

The [check-jsonschema](https://github.com/python-jsonschema/check-jsonschema) tool can be used to validate this, but there are many alternative tools that could be used:

```
❯ check-jsonschema --schemafile json-schema/group.json examples/example-group/zarr.json
Schema validation errors were encountered.
  examples/example-group/zarr.json::$: 'zarr_format' is a required property
```

Note that this only validates metadata stored within the `zarr.json` objects. It has no bearing on the actual data in the chunk files.

In addition to the schemas, I've included *the metadata* for a few examples, and have validated them against the json schema.

This is motivated by https://github.com/zarr-developers/geozarr-spec/issues/72. geozarr can define its own json schema for the additional properties it adds.


json-schema/array.json
```diff
@@ -0,0 +1,639 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "$id": "https://zarr-specs.readthedocs.io/v3/json-schema/array.json",
```
TomAugspurger, line 3: We'll need to find a suitable URI for the schema itself. Ideally we can publish the schema on tags to this repository through CI/CD.

STAC uses https://schemas.stacspec.org/<version>/item-spec/json-schema/<object-type>.json, for example https://schemas.stacspec.org/v1.1.0/item-spec/json-schema/item.json.

json-schema/array.json
```diff
@@ -0,0 +1,639 @@
...
+            "allOf": [
+                {
+                    "$ref": "#/$defs/extension_object"
+                },
+                {
+                    "properties": {
+                        "name": {
+                            "type": "string",
+                            "not": {
+                                "anyOf": [
```
TomAugspurger, line 521: This is a common pattern for all the extension objects (`data_type`, `codec`, etc.). We're using the [`oneOf`](https://json-schema.org/understanding-json-schema/reference/combining#oneOf) keyword to ensure that the data type, say, matches *exactly* one data type definition.

To ensure that a data type like `"bool"` doesn't match against both the core bool data type and an extension data type, we need to prohibit extension types from shadowing a core data type.

json-schema/group.json
```diff
@@ -0,0 +1,58 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "$id": "https://zarr-specs.readthedocs.io/v3/json-schema/group.json",
```
TomAugspurger, line 3: Also needs a permanent URI.

d-v-b:
this is awesome work tom!

jbms:
This is great work. 

It does seem rather unfortunate to have to list all of the ids defined in the core spec redundantly in order to exclude them as valid extension names. 

One idea would be to just pull in all of the schemas from zarr-extensions automatically (e.g. via a program that generates the schema), and disallow in the schema unknown IDs.

We could update zarr-extensions to include separate schemas for the core ids also.  That way almost everything could be pulled in just from zarr-extensions.

TomAugspurger:
> It does seem rather unfortunate to have to list all of the ids defined in the core spec redundantly in order to exclude them as valid extension names.

My natural preference is for simple / dumb solutions. In this case I'm probably fine with repeating the names since CI should immediately fail if we add some new core object but forget to update the list of fields. I *think* it's impossible for these to get out of sync. 

> We could update zarr-extensions to include separate schemas for the core ids also.

I wasn't aware of the `zarr-extensions` repo until after I submitted this PR. My initial preference would be to keep the JSON schema in the same repository otherwise it's (even more) likely to fall out of date as the spec evolves.

json-schema/array.json
pzaborowski: Good thing is if that is resolvable. You can also use this prefix if do not plan to keep own domain 'forever' https://schemas.opengis.net/. As you'll see here not all JSON schemas has Id (xml namespaces were more disciplined here) and these which has not always use the host e.g. https://schemas.opengis.net/os-geojson/1.0/example-1-eo-collections.json

json-schema/group.json
```diff
@@ -0,0 +1,58 @@
...
+    "title": "Zarr v3 Group Metadata Schema",
+    "description": "JSON Schema for Zarr v3 Group metadata documents.",
```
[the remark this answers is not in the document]
pzaborowski: looks like array.json schema could use $ref top level to this file not to replicate definitions. Resolvable URI thath can be changed later would help here, indeed.

jbms:
> > It does seem rather unfortunate to have to list all of the ids defined in the core spec redundantly in order to exclude them as valid extension names.
> 
> My natural preference is for simple / dumb solutions. In this case I'm probably fine with repeating the names since CI should immediately fail if we add some new core object but forget to update the list of fields. I *think* it's impossible for these to get out of sync. 

If something is missing from the list of exclusions then it will just also validate as an extension, meaning the configuration doesn't get checked. 

Additionally, if you make a typo in an identifier it will also just be considered an extension and validate successfully.

> 
> > We could update zarr-extensions to include separate schemas for the core ids also.
> 
> I wasn't aware of the `zarr-extensions` repo until after I submitted this PR. My initial preference would be to keep the JSON schema in the same repository otherwise it's (even more) likely to fall out of date as the spec evolves.

Putting the separate schemas in this repo instead would also be fine, or zarr-extensions could even be merged into this repo.
> 
> But it'd be good to figure out some way to share what's already be done there (CI / tooling maybe?) with what's proposed here. I'll take a closer look when I get a chance.

That repo basically has the complement of what you have here represented as a schema.

json-schema/array.json
```diff
@@ -0,0 +1,639 @@
...
+                    ],
+                    "properties": {
+                        "cname": {
+                            "type": "string",
+                            "enum": [
+                                "blosclz",
+                                "lz4",
+                                "lz4hc",
+                                "zlib",
+                                "zstd"
```
d-v-b, line 341: i think snappy is missing from this list

json-schema/array.json
```diff
@@ -0,0 +1,639 @@
...
+                    "const": "crc32c"
+                }
+            },
+            "additionalProperties": false
+        },
+        "storage_transformer": {
+            "$ref": "#/$defs/extension_object",
+            "description": "Storage transformer configuration"
+        },
+        "extension_object": {
```
d-v-b, line 488: I think this needs to be `string | {"name": string, "configuration"?: {...}}`. I don't know JSON schema well enough to see if the string variant is handled here.

d-v-b:
this is great @TomAugspurger, how can we get this out of draft?

TomAugspurger:
It's been a bit since I looked at this :)

It looks like there's [some discussion](https://github.com/zarr-developers/zarr-specs/pull/347#pullrequestreview-2898040430) about how to deduplicate field definitions across Group and Array. I don't know enough about jsonschema to comment.

Then there's the discussion about reserved fields: https://github.com/zarr-developers/zarr-specs/pull/347#issuecomment-2930731552.

Finally, we'll need to verify that the json schema (still) accurately reflects the text. I went through and manually verified things back when I opened this, but things may have drifted since then.

I think we can sort out stable URLs and how to publish these in a separate PR.

d-v-b:
let me know if you want any help! I would love to see this advance :)

PR #368, opened 2025-11-24
Clarify nesting. Add "configuration" as an index_location

mkitti:
# Abstract

This pull request enhances the sharding codec specification by introducing a new mechanism for storing the shard index and clarifying several related behaviors. Key changes include:

*   **Introduction of `configuration` index_location:** A new `configuration` value for `index_location` is introduced, allowing the shard index to be defined directly within the codecs configuration parameters in the array metadata.
*   **New `predefined_index` parameter:** When `index_location` is `configuration`, the `predefined_index` parameter specifies the shard index data, which can be either a JSON array or a BASE64 encoded string.
*   **Conditional `index_codecs` application:** Clarifications have been added on how `index_codecs` apply based on the type of `predefined_index`. If `predefined_index` is a BASE64 string, `index_codecs` are used to interpret the binary content. If its a JSON array, `index_codecs` may contain `array -> array` transformations but MUST NOT contain an `array -> bytes` codec.
*   **Read-Only Consideration:** Implementations MAY consider arrays using `index_location: configuration` (at any nesting level) to be read-only due to the complexities of updating such an index.
*   **Nested Shards Flexibility:** The specification now clarifies that nested shards MAY use the `configuration` value for `index_location`.
*   **Version Bump:** The version number is bumped to 1.1 to reflect these changes.

# Configuration Index

The `configuration` index location provides a mechanism to embed the shard index directly within the `sharding_indexed` codecs configuration object in the Zarr array metadata (`zarr.json`). This is particularly useful for pre-defined, fixed, or read-only datasets where the index does not change.

The index data is provided via the `predefined_index` parameter within the codec configuration. This parameter can take two forms:
*   A **JSON array**: This directly represents the decoded index (e.g., an array of `[offset, nbytes]` pairs). In this case, `index_codecs` may contain `array -> array` codecs to transform the conceptual `[N,2]` index to the JSON structure, but MUST NOT contain an `array -> bytes` codec.
*   A **BASE64 encoded string**: This represents the binary output of the `index_codecs` chain applied to the index array. Implementations should decode the BASE64 string and then apply the `index_codecs` in reverse to obtain the index.

# Shard Index as an Array

`index_codecs` was suggested to begin with a `bytes` array-to-bytes codec, implying that the shard index is an array. Clarifying that the shard is an array makes it clearer that array-to-array codecs could be used to manipulate the shard index format before its converted to bytes (for `start`/`end`/`BASE64` cases) or represented directly as a JSON array.

Potential applications:
*   The `transpose` codec could be used to have a consecutive offsets subarray followed by a consecutive nbytes subarray. This would more closely correspond with TIFFs `TILEOFFSETS` and `TILEBYTECOUNTS` tags.
*   Describing the distinct offset or nbytes subarrays an arithmetic sequence or repeated elements.
*   Shrinking the the shard index by casting the array from `uint64` to a smaller unsigned integer such as `uint32`, `uint16`, or `uint8`.

# Nested Shards

The specification now explicitly allows nested shards to utilize an `index_location` value of `configuration`. This provides greater flexibility in defining hierarchical data structures. However, due to the difficulty of updating an index stored in the array metadata, implementations MAY consider any array using `"index_location": "configuration"` (at any level of nesting) to be read-only. Writing to such an array may produce an error or lead to a corrupted state if the written data would require a change to the predefined index.

### DIFF
--- a/docs/v3/codecs/sharding-indexed/index.rst
+++ b/docs/v3/codecs/sharding-indexed/index.rst
@@ -147,27 +147,48 @@ Sharding can be configured per array in the :ref:`array-metadata` as follows::
 
 ``index_codecs``
 
-    Specifies a list of codecs to be used for encoding and decoding a shard index.
+    Specifies a list of codecs to be used for processing a shard index.
     The shard index is an array with a ``shape`` of ``[N,2]`` and a ``data_type`` of
     ``uint64`` where ``N`` is the number of chunks to be indexed in the shard.
     The ``index_codecs`` value must be an array of objects, as specified in the
-    :ref:`array-metadata`. The ``index_codecs`` member is required and needs to 
-    contain exactly one ``array -> bytes`` codec. That codec MAY be preceded by
-    ``array -> array`` codecs that modify either the ``shape`` or ``data_type``
-    of the array. Codecs that produce variable-sized encoded representation,
-    such as compression codecs, MUST NOT be used for index codecs. It is
-    RECOMMENDED to use a little-endian codec followed by a crc32c checksum as
-    index codecs.
+    :ref:`array-metadata`.
+
+    Unless ``index_location`` is ``configuration`` and `predefined_index` is a
+    JSON array, the ``index_codecs`` member is required and needs to contain
+    exactly one ``array -> bytes`` codec. This ``array -> bytes`` codec MAY be
+    preceded by ``array -> array`` codecs that modify either the ``shape`` or
+    ``data_type`` of the array.
+
+    If ``index_location`` is ``configuration`` and `predefined_index` is a JSON array,
+    the ``index_codecs`` member MAY be empty or contain only ``array -> array``
+    codecs. An ``array -> bytes`` codec MUST NOT be present. The ``array -> array``
+    codecs define the transformation from the conceptual `[N, 2]` index to the
+    structure represented in the JSON array.
+
+    Codecs that produce variable-sized encoded representation, such as
+    compression codecs, MUST NOT be used for index codecs. It is RECOMMENDED
+    to use a little-endian codec followed by a crc32c checksum as index codecs
+    when an ``array -> bytes`` codec is used.
 
 ``index_location``
 
     Specifies whether the shard index is located at the beginning of the file,
-    the end of the file, or external to the file in its own key. The parameter
-    value must be either the string ``start``, ``end``, or ``external``. A value
-    of external indicates the shard index has been written to an external file
-    referred to by a key constructed by appending ".shard_index" to the key of
-    the sharded chunk.  If the parameter is not present, the value defaults to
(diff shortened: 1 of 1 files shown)


jbms:
I can see that in some cases this format would be useful.  However this proposal is

jbms:
I can see that a separate file may be useful in some cases.  However that is not possible under the existing codec model.  We would need to define something more general, e.g. an array -> files codec.  That is something I mentioned before in the context of making the chunk_grid itself a part of the codec pipeline rather than a separate higher level thing.

mkitti:
Could we implement the external shard indices via a storage transformer?

The storage transformer looks for `.shard_index` files and then prepends or appends them to the byte values of the corresponding key?

mkitti:
> Could we implement the external shard indices via a storage transformer?
> 
> The storage transformer looks for `.shard_index` files and then prepends or appends them to the byte values of the corresponding key?

I think this might work for reading, but a storage transformer might need additional information for writing.

mkitti:
I proposed the storage transformer `concat-parts` as a zarr-extension pull request:
https://github.com/zarr-developers/zarr-extensions/pull/39

mkitti:
I have removed the "external" index_location from this pull request.

Instead I have a proposed a "configuration" index_location where the shard index is defined in the codec configuration as a "predefined_index" parameter. At the moment, I have written that the parameter may either be a JSON array itself or a BASE64 encoded string.

Allowing the "predefined_index" parameter to be a JSON array adds some complications. How should `index_codecs` be interpreted in that case? Here I specified that `index_codecs` in that case should not end in conversion to bytes but either be empty or only contain array-to-array codecs.

LDeakin:
Could you clarify the purpose of `"predefined_index"`. For some other file type to masquerade as a zarr chunk?

If every single shard has the same shard index then:
- Subchunks in each shard must be at the same offset/size, so no compression is feasible and the order is constrained
- The only compatible chunk grid is `regular`, because you can't have shards with a different shape

mkitti:
> Could you clarify the purpose of `"predefined_index"`. For some other file type to masquerade as a zarr chunk?
> 
> If every single shard has the same shard index then:
> - Subchunks in each shard must be at the same offset/size, so no compression is feasible and the order is constrained
> - The only compatible chunk grid is `regular`, because you can't have shards with a different shape

Mapping an existing foreign file as a Zarr shard is one purpose. See the various virtual Zarr efforts.

I disagree that "no compression" is feasible. We might need some padding or offset codec though to fit everything into a grid.

mkitti:
@LDeakin , I did initially consider external indices here as well but @jbms convinced me that does not really work as a codec.

Rather I proposed this as a storage transformer where you can separate a "value" into a separate keys with extensions:
https://github.com/zarr-developers/zarr-extensions/pull/39

Issue #371, opened 2026-01-02
zarr-python's consolidated metadata implementation violates the spec


d-v-b:
[Previously](https://github.com/zarr-developers/zarr-specs/blob/350d87e6b55b1ab3440308b9533d28d33892d72e/docs/v3/core/v3.0.rst?plain=1#L688-L692), additional fields in metadata were allowed as long as they were JSON objects with a `must_understand: false` key: value pair. Zarr-python's consolidated metadata implementation complied with this requirement (see script below). 

The recent redefinition of extra fields in metadata documents added the [requirement](https://github.com/zarr-developers/zarr-specs/blob/dc3e95ed36060d9533361364ab7f54fe3e53f82b/docs/v3/core/index.rst?plain=1#L1528-L1531) that such extra fields have a `name` key which is a string. zarr-python's consolidated metadata does not contain a `name` key, and so it is out of spec.

As consolidated metadata is used heavily by xarray users, this is a very high-impact change. The recent spec refactor has thus made many  (most?) zarr v3 xarray datasets technically out of spec.

```python
# /// script
# dependencies = [
#   "zarr==3.1.0",
# ]
# ///

import zarr
from pprint import pprint
import json

store = {}
zarr.create_group(store)
consolidated = zarr.consolidate_metadata(store)
pprint(json.loads(store["zarr.json"].to_bytes()))
"""
{'attributes': {},
 'consolidated_metadata': {'kind': 'inline',
                           'metadata': {},
                           'must_understand': False},
 'node_type': 'group',
 'zarr_format': 3}
"""
``` 

I think we should treat this as a regression in the spec. A fix could be:
- clarify that readers may ignore _any_ additional field that is a JSON object with a `must_understand: false` key value pair, no matter what other keys that object has.
- remove the requirement that top-level extra fields in array / group metadata objects have a `name` field if they are JSON objects. 

Without these two changes, or changes that achieve the same effect, a large volume of zarr data is out of spec, and we need to fix that.

d-v-b:
I'm working on something that could conceivably be an additional, `must_understand: False` array metadata field, so it would be really helpful to know if this is actually permitted by the spec!

maxrjones:
I wanted a bit more context around this change but find it difficult to track conversations across the ZEPS/specs, so I had claude create this detailed context document - https://hackmd.io/8cBgorBeR4Sv5hrZ5g5NSw. This approach has issues (including overconfident assessments), but did help me get started thinking about this.

[...]

PR #373, opened 2026-01-10
adjust language about extra keys to permit consolidated metadata

d-v-b:


docs/v3/core/index.rst
```diff
@@ -777,13 +775,10 @@ Optional keys:
     pairs, where the key must be a string and the value can be an arbitrary
     JSON literal. Intended to allow storage of arbitrary user metadata.
 
-.. _group-metadata-extensions:
-
-Unknown
```
d-v-b, line 782: "Unknown" isn't apt, since extra keys can be contrary to the spec even if they are "known"

docs/v3/core/index.rst
```diff
@@ -777,13 +775,10 @@ Optional keys:
...
-^^^^^^^
+Additional fields
+^^^^^^^^^^^^^^^^^
 
-All other keys found in the metadata object MUST be interpreted
-following the :ref:`Extensions section <extensions_section>`.
+See :ref:`Additional fields<additional_fields>`.
```
d-v-b, line 781: the main effect of this PR is to decouple the definition of an extension from requirements for extra fields. To be honest I don't find "extension" definition is very useful -- it's more natural to talk about codecs, data types, etc rather than lumping them all into one category.

docs/v3/core/index.rst
```diff
@@ -799,6 +794,15 @@ For example, the JSON document below defines a group::
...
     }
 
+.. _additional_fields:
+
+Additional fields
+=================
+
+If a reader encounters a Zarr array or group metadata document bearing a field not defined in this 
+document, the reader MUST fail to open the array or group unless the additional field is a 
+``JSON`` object containing a ``"must_understand": false`` key-value pair.
```
d-v-b: for reference, here is the old text: 

https://github.com/zarr-developers/zarr-specs/blob/350d87e6b55b1ab3440308b9533d28d33892d72e/docs/v3/core/v3.0.rst?plain=1#L688-L692

d-v-b: I tried to capture the (contradictory) requirement that no additional fields be present, and that additional fields are OK if they contain `"must_understand": false`

d-v-b:
closes #371

d-v-b:
ping @rabernat for review

d-v-b:
consolidated metadata is now permitted as an explicit exception to the new general rule. This keeps the intent of the 3.1 changes, but keeps extant data in-spec.

docs/v3/core/index.rst
```diff
@@ -799,6 +794,30 @@ For example, the JSON document below defines a group::
...
+        "must_understand": false,
+        "kind": "inline",
+        "metadata": {...}
+    }
+
+Note that this object lacks the ``name`` member required of extension objects and contains the
+additional ``kind`` and ``metadata`` members.
+
+The use of ``consolidated_metadata`` as described above is widespread, but writers should avoid it in favor
+of alternatives that preserve the standard field structure of the group metadata document.
```
maxrjones: I don't think the recommendation to avoid it is helpful unless it comes with a specific, working example of the alternative option.

d-v-b: yes, we need an attributes-based convention. I am happy to draft one, if we get buy-in from implementations in the blast radius. I think we can speak for zarr-python in saying that we would be happy to add support for a new implementation of consolidated metadata.

maxrjones: I agree we need a long term solution. I also think this problem can be broken into two parts:

1. A fix for the fact that implementations are not following the letter of the spec since https://github.com/zarr-developers/zarr-specs/blob/dc3e95ed36060d9533361364ab7f54fe3e53f82b/docs/v3/core/index.rst?plain=1#L1528-L1531 (xref https://github.com/zarr-developers/zarr-specs/issues/371)
2. A long-term solution for reducing metadata requests 

I suggest that this PR only tries to solve the first component of the current predicament so that its resolution is not prolonged by the design discussions and implementation coordination required for the second part of the problem.

For this specific sentence, my recommendation would be to just remove it and add it later when a better solution exists. I think this is similar to some of the conversations we've had in zarr-python about regretting putting empty stubs for future features and how it's better to add the pointer to the new feature after they already exist.

d-v-b: I'm happy to remove that sentence, with the goal of adding a pointer that guides people to a recommended solution when we develop one

d-v-b: resolved in https://github.com/zarr-developers/zarr-specs/pull/373/commits/094ab4e8568a6add02bc0c98d244b33715170d72

d-v-b:
@LDeakin @jbms would you mind having a look? I think this PR, or something functionally equivalent, is necessary to keep consolidated metadata consistent with the spec.

LDeakin (approved):
Just jotting down my thoughts here:
1. As noted in https://github.com/zarr-developers/zarr-specs/issues/371, the spec regressed in 3.1 with enforcing the extension definition for additional fields[^1]
2. It seems like `consolidated_metadata` is the only additional field written by implementations that is non-conformant with that definition (and the only additional field outright).

Given (1) and (2), I think this regression is acceptable with the added text of this PR. It doesn't lock us in to doing anything new/changing consolidated metadata either. Consolidated metadata is so prevalent that any change (e.g. move to `attributes`/ZEP10 style generic extensions) will have to be coordinated across implementations, maintaining backwards compatibility on read.

I'd also flag that the alternative to this PR is to fully reverse the regression, but I'd rather we move forward with ZEP10 and avoid adding any future _generic extensions_ as additional top-level fields (beyond `extensions`).

[^1]: I did not change `zarrs` after this regression. `zarrs` continues to support [3.0 style additional fields](https://docs.rs/zarrs_metadata/latest/zarrs_metadata/v3/struct.AdditionalFieldV3.html) and `consolidated_metadata` is parsed off that rather than being specially handled in [`GroupMetadataV3`](https://docs.rs/zarrs_metadata/latest/zarrs_metadata/v3/struct.GroupMetadataV3.html)

normanrz:
I agree with @LDeakin here. It makes sense to have a *hotfix* for consolidated_metadata and work towards a more general extension system (ZEP10) separately.

PR #375, opened 2026-03-17
no nodes named zarr.json

### LINKED ISSUE #374: disallow arrays or groups named "zarr.json"
Practically, there are 2 ways to identify an array or group: by name, or by `<name>/zarr.json`. The former is the "hierarchy name", and the latter is the actual storage key that defines the presence of the array or group. 

The hierarchy name is generally not a stored object, but rather a prefix. On local storage, this means `<name>` is a directory, and on object storage, `<name>` does not resolve to anything, it has no content type, etc. By contrast, `<name>/zarr.json` is a stored object.

This means there are some advantages to identifying a Zarr array or group by its stored representation, i.e. `<name>/zarr,json`:
- As a file, zarr,json can easily support drag + drop behavior for GUI applications that work with conventional single-file data formats.
- as a resource, zarr.json can be associated with a JSON schema and a content type.

However, to support these two forms of declaration (`<name>` or `<name>/zarr.json`), a client would need a way to distinguish them structurally. Is `<name>/zarr.json` the name of a group called `zarr.json` (i.e., should the client check if `<name>/zarr.json/zarr.json` exists?), or the key of a metadata document for the group called `<name>`? On object storage _both_ could even be true -- `zarr.json`  and `zarr.json/zarr.json` can both be the keys for metadata documents.

To allow clients to treat `<name>` and `<name>/zarr.json` as valid identifiers for the array or group `<name>`, we need to exclude the possibility that `<name>/zarr.json` is the name of an array or group. IMO the best way to do this is by stipulating that arrays and groups MUST not be named `zarr.json`. If we add this restriction, then users can safely pass strings like `foo` OR `foo/zarr.json` into zarr-aware applications without ambiguity, and Zarr data can be declared in an object-stored-friendly way.

d-v-b:
closes #374

### DIFF
--- a/docs/v3/core/index.rst
+++ b/docs/v3/core/index.rst
@@ -811,6 +811,7 @@ apply to node names:
 * must not include the character ``"/"``
 * must not be a string composed only of period characters, e.g. ``"."`` or ``".."``
 * must not start with the reserved prefix ``"__"``
+* must not be ``"zarr.json"``
 
 To ensure consistent behaviour across different storage systems and programming
 languages, we recommend users to only use characters in the sets ``a-z``,


lubianat (approved):
I have limited experience, but this looks like a clear clean up

jbms (approved):

LDeakin (approved):

mkitti (approved):
Technically breaking, but I'm assuming that if anyone named a group or array "zarr.json" they were asking for trouble.

PR #330, opened 2025-02-14
ZEP9 (phase 1): add clarifications for extension naming

joshmoore:
This PR clarifies the extension mechanism concept in the v3 specification. Comments on any changes which will break existing implementations are **STRONGLY** encouraged. Please see https://github.com/zarr-developers/zeps/pull/65 for background material.

TODOs:
 * [x] clarify the file numbering (currently 3.0.rst)
 * [x] move definitions to the appropriate location (core, subtype page or [ext](https://github.com/zarr-developers/zarr-extensions) 

Post-merge:
 * [ ] remove preview label from [zarr-extensions](https://github.com/zarr-developers/zarr-extensions) 
 * [ ] return to https://github.com/zarr-developers/zarr-specs/issues/329
 * [ ] update sphinx to clearly show 3.0 vs 3.1

PR #393, opened 2026-09-16
core: allow the extension-object form for core data types

### LINKED ISSUE #392: JSON object form for core data types
`{"name": "int8"}` is invalid according to [this statement](https://github.com/zarr-developers/zarr-specs/blame/fc7dd9c9beb5a50b87f9b08b00bf50fc0048482f/docs/v3/core/index.rst#L509-L512):

> If the data type is defined in :ref:`this specification <data-type-list>`,
    then the value must be the data type identifier provided as a string. For example, ``"float64"`` for
    little-endian 64-bit floating point number.

This restriction, dating from the 3.0 release of the spec, conflicts with the newer [language about extensions](https://github.com/zarr-developers/zarr-specs/blame/fc7dd9c9beb5a50b87f9b08b00bf50fc0048482f/docs/v3/core/index.rst#L1542-L1548):

> In `metadata documents`_, extensions can be encoded either as objects or as
short-hand names.
> 
> If using an object definition, the member ``name``
MUST be a plain string which conforms to :ref:`extension name <extension-naming>`.
Optionally, the member ``configuration`` MAY be present but if so MUST be
an object.

I think we should aim for consistency here. To me, consistency argues for removing the requirement that core data types be declared only by strings, and allowing all the various forms introduced by the 3.1 changes:
- `<name>`
- `{"name": <name>}`
- `{"name": <name>, "configuration": {}}`
- `{"name": <name>, "configuration": {}, "must_understand": true}`
- `{"name": <name>, "must_understand": true}`

(I think that's all of them 😆 )

d-v-b:
This is a small change that widens the acceptable JSON forms for data types to include JSON object variants. The edits were a mix of claude and me.

:robot: _AI text below_ :robot:

Closes #392.

The `data_type` section still says core data types "must be the data type identifier provided as a string". That sentence dates from 3.0 (`405eb2b2`), when core types were strings and extension types were objects. The 3.1 extension rewrite (#330) made `data_type` an extension point that "MUST conform to the extension-definition", where short-hand names and objects are equivalent — but the string-only sentence was kept (it only gained a `:ref:` link in `b24440e8`), so `{"name": "int8"}` is valid by the extension definition and invalid by the `data_type` section.

This PR resolves the conflict in the direction #392 proposes: core data types may use any form the extension definition permits. The core identifier becomes the extension `name`, whether written as a short-hand name or inside an object.

Also dropped "little-endian" from the `float64` example — in v3 byte order is the `bytes` codec's concern, not the data type's.

No change-log entry, following #373; happy to add one if wanted.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


LDeakin:
Should this clarify here that data types must not have a false "must_understand"?

d-v-b:
Yes it should.

LDeakin (approved):

